PR #6, opened 2022-10-11
SW-4007: State default argument in the ros launch files for optional ones

Samahu:
## Related Issues & PRs
- N/A

## Summary of Changes
- Adjust argument parsing such that it is known which args are required and which are optional
- Allow the user to pass `timestamp_mode` during replay to override the timestamp on IMU and LIDAR packets. 

## Validation
* The following command should reflect the exact requirement of what is considered a required roslaunch file argument as specified in the [usage section](https://github.com/ouster-lidar/ouster-ros#usage) of the repo readme file
```bash
roslaunch ouster_ros sensor.launch --ros-args
roslaunch ouster_ros record.launch --ros-args
roslaunch ouster_ros replay.launch --ros-args
```
* Invoking `roslaunch ouster_ros sensor.launch` without specifying `sensor_hostname` should fail immediately
* Invoking `record.launch` or `replay.launch` without specifying `metadata` should fail immediately.
* It is now possible to specify the `timestamp_mode` parameter value during replay.
  - Setting `timestamp_mode` to **TIME_FROM_ROS_TIME** modifies the timestamp of published IMU and Lidar ROS messages.  
* Functionality is maintained

### DIFF
--- a/include/ouster_ros/os_client_base_nodelet.h
+++ b/include/ouster_ros/os_client_base_nodelet.h
@@ -19,6 +19,10 @@ class OusterClientBase : public nodelet::Nodelet {
     virtual void onInit() override;
 
    protected:
+    bool is_arg_set(const std::string& arg) {
+        return arg.find_first_not_of(' ') != std::string::npos;
+    }
+
     void display_lidar_info(const ouster::sensor::sensor_info& info);
 
    protected:
--- a/src/os_sensor_nodelet.cpp
+++ b/src/os_sensor_nodelet.cpp
@@ -79,7 +79,7 @@ class OusterSensor : public OusterClientBase {
 
     void save_metadata(ros::NodeHandle& nh) {
         auto meta_file = nh.param("metadata", std::string{});
-        if (!meta_file.size()) {
+        if (is_arg_set(meta_file)) {
             meta_file =
                 hostname.substr(0, hostname.rfind('.')) + "-metadata.json";
             NODELET_WARN_STREAM(
@@ -141,7 +141,7 @@ class OusterSensor : public OusterClientBase {
     std::shared_ptr<sensor::client> create_client(const std::string& hostname,
                                                   int lidar_port,
                                                   int imu_port) {
-        if (!hostname.size()) {
+        if (hostname.empty()) {
             auto error_msg = "Must specify a sensor hostname";
             NODELET_ERROR_STREAM(error_msg);
             throw std::runtime_error(error_msg);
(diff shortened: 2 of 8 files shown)


include/ouster_ros/os_client_base_nodelet.h
```diff
@@ -19,6 +19,10 @@ class OusterClientBase : public nodelet::Nodelet {
     virtual void onInit() override;
 
    protected:
+    bool is_arg_set(const std::string& arg) {
+        return arg.find_first_not_of(' ') != std::string::npos;
```
kairenw, line 23: for my edification :joy: what makes this approach superior to using `""` and `.empty()` like we were before

src/os_sensor_nodelet.cpp
```diff
@@ -167,16 +167,14 @@ class OusterSensor : public OusterClientBase {
...
-        auto lidar_mode_arg = nh.param("lidar_mode", std::string{});
-        auto timestamp_mode_arg = nh.param("timestamp_mode", std::string{});
         auto lidar_port = nh.param("lidar_port", 0);
         auto imu_port = nh.param("imu_port", 0);
-
-        std::string udp_profile_lidar_arg;
-        nh.param<std::string>("udp_profile_lidar", udp_profile_lidar_arg, "");
+        auto lidar_mode_arg = nh.param("lidar_mode", std::string{});
+        auto timestamp_mode_arg = nh.param("timestamp_mode", std::string{});
+        auto udp_profile_lidar_arg = nh.param("udp_profile_lidar", std::string{});
```
kairenw, line 174: but we don't want `" "` here?

include/ouster_ros/os_client_base_nodelet.h
Samahu: It is not superior, it was rather needed. When you use `default=""` attribute in the **arg** definition it is still rendered as if you didn't initialize the argument and it shows this as a required value. Instead I used `default=" "` which includes a space. Hence I needed to use `arg.find_first_not_of(' ') != std::string::npos` to check if an arg has been initilaized.

src/os_sensor_nodelet.cpp
Samahu: I didn't understand the question

include/ouster_ros/os_client_base_nodelet.h
Samahu: An unuset arg is left unchanged during initialized. The alternative is to either expose and use `sensor::*_UNSPEC` to the launch files which in turn would leave the config value unchanged during sensor initialization, or configure each argument with an already exposed value such as `sensor::MODE_1024x10` as the default value but then the user would need to change the value whenever they want to launch the node with different values or edit the files.

Long story short I went with the white space to indicate that the value is unchanged but I am open to suggestions, specifically the two other alternatives that I pointed out.

kairenw: Ah, I understand. So now we get the correct error instead of how it used to fail because it couldn't resolve. 
Happy with your method. Don't really want to expose `*_UNSPEC` or force users to change values every time

kairenw (approved):

src/os_sensor_nodelet.cpp
kairenw: haha don't worry about it

PR #24, opened 2022-11-10
Declare missing curl dependency

mortenfyhn:
## Related Issues & PRs

None.

## Summary of Changes

This declares the curl dependency in the package manifest. On most systems, curl is already installed, so this doesn't make any difference, but if you're building this on a minimal system (docker container, for instance), then you'll have a broken build.

## Validation

I've built the package with and without declaring this, in a docker container that doesn't have curl preinstalled. When declared, rosdep picks it up and all is well.


Samahu (approved):
Looks good to me

PR #43, opened 2022-12-14
Change informational prints from WARN to INFO

mortenfyhn:
Hi!

The code has some WARN level prints that really aren't warnings. I've changed these to INFO level prints.

I'm guessing they were made WARN so that they are easier to find in a long list of output, but I still think they're better off as INFO level statements. Having warnings in the log makes me think something is wrong and needs to be fixed. My brain automatically parses orange text as something bad, so I reckon it's best to avoid orange warnings when everything is good.

If you agree, feel free to merge.


Samahu (approved):
I do agree with you, we shouldn't be utilize WARN to display INFO messages. Approved

mortenfyhn:
Lovely, thank you!

PR #53, opened 2023-01-27
ROS-53: Add parameter for utc/tai offset in ptp mode.

jackg0:
## Summary of Changes

Add `ptp_utc_tai_offset_secs` parameter to account for UTC/TAI offset.

International Atomic Time (TAI) is currently ahead of UTC by 37 seconds, so that is the default setting for the new parameter.

## Validation

- Tested with OS0 32 beam lidar on robot with hardware ptp support with ptp4l and phc2sys running as services.

- Used the following command to launch the driver:

```
roslaunch ouster_ros sensor.launch sensor_hostname:=<sensor-ip> viz:=false timestamp_mode:=TIME_FROM_PTP_1588 lidar_mode:=1024x10 ptp_utc_tai_offset_secs:=-37
```

- Previously, timestamps for the point clouds published on `ouster/points` were in the future by approximately -36.9 seconds compared to UTC system time. After the change, the timestamps are now milliseconds in the past as expected. Similarly for the imu messages published on `ouster/imu`, the timestamp difference is now less than a millisecond compared with system time.

### DIFF
--- a/launch/common.launch
+++ b/launch/common.launch
@@ -5,6 +5,7 @@
   <arg name="rviz_config" doc="optional rviz config file"/>
   <arg name="tf_prefix" doc="namespace for tf transforms"/>
   <arg name="timestamp_mode" doc="method used to timestamp measurements"/>
+  <arg name="ptp_utc_tai_offset_secs" doc="UTC/TAI offset to apply when using TIME_FROM_PTP_1588"/>
 
   <group ns="$(arg ouster_ns)">
     <node pkg="nodelet" type="nodelet" name="os_cloud_node"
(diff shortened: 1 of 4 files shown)


doisyg:
I have the same issue. Why was this closed ?

Samahu:
> @Samahu It is likely that new utc-tai leap seconds will be introduced in the future. The offset will have to be changed manually. Doesn't the PTP grandmaster already communicate the current offset between UTC and TAI? So can't the UTC be computed instead of configured? It's not within the scope of this project, so I think it is nice that it can be configured here, but wouldn't it be better if the Ouster itself is able to compute the offset and take it into account, so that the ROS package would be agnostic to this? I think this is how it's usually done with other sensors as well.

I am not very familiar on how the UTC-TAI offset can be computed automatically. This would be a low priority item on my list but I would consider any contribution that can implement this (with the ability to manually override the offset).

RonaldEnsing:
> I am not very familiar on how the UTC-TAI offset can be computed automatically. This would be a low priority item on my list but I would consider any contribution that can implement this (with the ability to manually override the offset).

I think this has to be implemented on the code running on the Ouster itself, not on the ROS2 driver. So it would not be possible to contribute this.

outrider-jkoch:
I would agree with @RonaldEnsing that this would need to happen in the firmware. However, because end users have already developed software that expects it to be a certain way if it were implemented it should probably be a user selectable option.

Samahu:
Thanks for the feedback, I will communicate this to our firmware product team

PR #64, opened 2023-02-22
Quick fixes in ROS2 - Foxy 

adyczech:
## Summary of Changes
Minor fixes of some issues that merged during testing of ros2-foxy branch.

- Fixed missing .cpp file extension in ouster-ros CMakeLists.txt
- Fixed arguments of `static_transform_publishers` in `rviz.launch.py`
- Definition of `wait_time_per_attempt` was moved to `ouster-ros/src/os_processing_node_base.cpp` since it was causing undefined symbol error while loading library os_image_component and os_cloud_component library
- Publishers `lidar_packet_pub` and `imu_packet_pub` were converted to LifecyclePublishers and activated in `on_activate` callback. Otherwise I was getting `Trying to publish message on the topic 'XXX', but the publisher is not activated`

## Validation
- Tested on Ubuntu 20.04, ROS2 foxy and Ouster OS0-128-U with FW 2.3.1 launching `sensor.independent.launch.py` since `sensor.launch.xml` points to `sensor.composite.launch.xml` that contains `node_container` which is not supported in foxy.

### DIFF
--- a/ouster-ros/CMakeLists.txt
+++ b/ouster-ros/CMakeLists.txt
@@ -143,7 +143,7 @@ rclcpp_components_register_node(os_cloud_component
 
 # ==== os_image_component ====
 create_ros2_component(os_image_component
-  "src/os_processing_node_base;src/os_image_node.cpp"
+  "src/os_processing_node_base.cpp;src/os_image_node.cpp"
   ""
 )
 rclcpp_components_register_node(os_image_component
--- a/ouster-ros/src/os_sensor_node.cpp
+++ b/ouster-ros/src/os_sensor_node.cpp
@@ -86,7 +86,9 @@ class OusterSensor : public OusterSensorNodeBase {
         const rclcpp_lifecycle::State& state) {
         RCLCPP_DEBUG(get_logger(), "on_activate() is called.");
         LifecycleNode::on_activate(state);
-
+        lidar_packet_pub->on_activate();
+        imu_packet_pub->on_activate();
+       
         allocate_buffers();
         if (!connection_loop_timer) {
             // TOOD: replace with a thread instead?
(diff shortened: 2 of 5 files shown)


ouster-ros/src/os_sensor_node.cpp
```diff
@@ -86,7 +86,9 @@ class OusterSensor : public OusterSensorNodeBase {
         const rclcpp_lifecycle::State& state) {
         RCLCPP_DEBUG(get_logger(), "on_activate() is called.");
         LifecycleNode::on_activate(state);
-
+        lidar_packet_pub->on_activate();
+        imu_packet_pub->on_activate();
+
```
Samahu, line 91: Isn't the `LifecycleNode::on_activate` doing the activation automatically?

ouster-ros/CMakeLists.txt
```diff
@@ -143,7 +143,7 @@ rclcpp_components_register_node(os_cloud_component
 
 # ==== os_image_component ====
 create_ros2_component(os_image_component
-  "src/os_processing_node_base;src/os_image_node.cpp"
+  "src/os_processing_node_base.cpp;src/os_image_node.cpp"
```
Samahu, line 146: how did I miss that? :eyes: :+1:

Samahu (approved):
@adyczech I did test your changes and you are right about the sensor.composite.launch.xml not working due to **node_container**.

ouster-ros/src/os_sensor_node.cpp
Samahu: It seems this is a necessary step.

Samahu:
@adyczech I went ahead and merged your fixes. I have also switched all of xml launch files to use the independent configuration by default in a separate commit.

PR #81, opened 2023-03-13
ROS1 - Adding latched topic transport config/metadata

TobinHall:
## Related Issues & PRs
Implements changes for #80 for ros1.
## Summary of Changes
Add a std_msgs/String latched publisher to os_sensor_nodelet.cpp.

In get_metadata() for both os_cloud_nodelet.cpp and os_image_nodelet.cpp:
- Add a std_msgs/String callback and subscriber.
- Manually spin until a service client becomes available or a non-empty config message has arrived. 

## Validation

1. Launched sensor.launch with a live sensor, confirmed that the metadata source is from topic. Confirmed that point-clouds and images are published. 
2. Launched sensor.launch with config topic remapped (making topic unavailable) , confirmed that the metadata source is not from topic (from service). Confirmed that point-clouds and images are published.

### DIFF
--- a/src/os_image_nodelet.cpp
+++ b/src/os_image_nodelet.cpp
@@ -34,7 +34,6 @@
 #include <vector>
 
 #include "ouster/image_processing.h"
-#include "ouster_ros/GetMetadata.h"
 
 namespace sensor = ouster::sensor;
 namespace viz = ouster::viz;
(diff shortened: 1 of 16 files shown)


src/os_sensor_nodelet.cpp
```diff
@@ -409,6 +414,8 @@ class OusterSensor : public OusterClientBase {
     void create_publishers(ros::NodeHandle& nh) {
         lidar_packet_pub = nh.advertise<PacketMsg>("lidar_packets", 1280);
         imu_packet_pub = nh.advertise<PacketMsg>("imu_packets", 100);
+        config_pub = nh.advertise<std_msgs::String>("config", 1, true);
```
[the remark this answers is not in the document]
Samahu: additionally, I think it would be better if we create this publisher very early on in the node.. probably right after we create the client and before we invoke the `update_config_and_metadata`. This publisher serves a different purpose than the two other publishers. This configuration the two other publishers are sending data.

src/os_sensor_nodelet.cpp
```diff
@@ -449,8 +456,10 @@ class OusterSensor : public OusterClientBase {
    private:
     PacketMsg lidar_packet;
     PacketMsg imu_packet;
+    std_msgs::String config_msg;
```
Samahu, line 459: If you follow my advice where I suggest to move the `metadata_pub` then we don't need to have this member variable.
```suggestion
    std_msgs::String config_msg;
```

src/os_sensor_nodelet.cpp
```diff
@@ -409,6 +414,8 @@ class OusterSensor : public OusterClientBase {
     void create_publishers(ros::NodeHandle& nh) {
         lidar_packet_pub = nh.advertise<PacketMsg>("lidar_packets", 1280);
         imu_packet_pub = nh.advertise<PacketMsg>("imu_packets", 100);
+        config_pub = nh.advertise<std_msgs::String>("config", 1, true);
+        config_pub.publish(config_msg);
```
Samahu, line 418: Not needed, if you follow my advice about moving the `metadata_pub` then we don't need this step since the block you added within `update_config_and_metadata` will publish the metadata as soon as we get results from the sensor. 
```suggestion
        config_pub.publish(config_msg);
```

src/os_image_nodelet.cpp
```diff
@@ -55,9 +56,29 @@ class OusterImage : public nodelet::Nodelet {
     }
 
     std::string get_metadata(ros::NodeHandle& nh) {
+        std::string config;
```
Samahu, line 59: rename to metadata
```suggestion
        std::string metadata;
```

src/os_image_nodelet.cpp
```diff
@@ -55,9 +56,29 @@ class OusterImage : public nodelet::Nodelet {
...
+        // Setup subscriber to get config via topic
+        auto config_handler = [&](const std_msgs::String::ConstPtr& p) {
+            config = p->data;
+        };
+        auto config_sub = nh.subscribe<std_msgs::String>("config", 1, config_handler);
+  
+        // Setup client to get config via service
         ouster_ros::GetMetadata request;
         auto client = nh.serviceClient<ouster_ros::GetMetadata>("get_metadata");
```
Samahu, line 68: I think we can do with the `ServiceClient` altogether?

src/os_image_nodelet.cpp
```diff
@@ -55,9 +56,29 @@ class OusterImage : public nodelet::Nodelet {
...
-        client.waitForExistence();
+  
+        // Wait for a valid config (via topic) or a valid service client
+        while(config.empty() && !client.exists()){
+            NODELET_WARN_THROTTLE(10.0, "OusterImage: waiting for config...");
+            ros::spinOnce();
+            ros::Duration(0.1).sleep();
+        }
```
Samahu, line 75: If we get right of the client then we need to add while a check to make sure ros::ok or wasn't not shutdown while we are waiting for the metadata to come in.

Samahu (changes requested):
This looks great, I just had few minor comments.

Samahu (approved):
@TobinHall I have made some improvements and extensions to your work and I am done reviewing your PR :tada:

PR #101, opened 2023-04-13
SW-4840: add a no-bond option to the ouster driver

Samahu:
## Related Issues & PRs
N/A

## Summary of Changes
- added a no-bond option to the ``sensor.launch`` file

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -19,7 +19,7 @@ ouster_ros(1)
   overflow
 * added a new ``/ouster/metadata`` topic that is consumed by os_cloud and os_image nodelets and
   save it to the bag file on record
-* make specifying metadata file optinal as of package version 8.1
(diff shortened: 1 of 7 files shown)


twslankard (approved):

kubelvla:
@Samahu Hi, the no_bond arg is required by common.launch, but not set in the replay.launch, so it fails to launch at all. Also, what is the preferred default value?

Samahu:
> @Samahu Hi, the no_bond arg is required by common.launch, but not set in the replay.launch, so it fails to launch at all.

Hi @kubelvla, thanks for reporting, I have amended the master branch to have the default for `no_bond` within `common.launch`. Please pull master branch again, sorry for the inconvenience.

On general, you should keep the `no_bond` set to false. You would only want to enable this flag when facing a `broke bond` errors. Which happens on a not well synchronized or congested networks.

PR #103, opened 2023-04-13
SW-4837: replace the use of ros service to retrieve sensor metadata with latched topics foxy

Samahu:
## Related Issues & PRs
- Related #102 

## Summary of Changes
* added a new `/ouster/metadata` topic that is consumed by os_cloud and os_image nodes and save it to the bag file on record
* make specifying metadata file optional during record and replay modes as of package version 8.1

## Validation
all functionality maintained

### DIFF
--- a/ouster-ros/src/os_replay_node.cpp
+++ b/ouster-ros/src/os_replay_node.cpp
@@ -30,7 +30,9 @@ class OusterReplay : public OusterSensorNodeBase {
 
         try {
             auto meta_file = parse_parameters();
-            populate_metadata_from_json(meta_file);
+            create_metadata_publisher();
+            load_metadata_from_file(meta_file);
+            publish_metadata();
             create_get_metadata_service();
             RCLCPP_INFO(get_logger(), "Running in replay mode");
         } catch (const std::exception& ex) {
@@ -110,7 +112,7 @@ class OusterReplay : public OusterSensorNodeBase {
         return meta_file;
     }
 
-    void populate_metadata_from_json(const std::string& meta_file) {
+    void load_metadata_from_file(const std::string& meta_file) {
         try {
             std::ifstream in_file(meta_file);
             std::stringstream buffer;
@@ -129,7 +131,6 @@ class OusterReplay : public OusterSensorNodeBase {
     void cleanup() {
         get_metadata_srv.reset();
     }
-
 };
 
 }  // namespace ouster_ros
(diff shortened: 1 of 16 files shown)


ouster-ros/src/os_replay_node.cpp
```diff
@@ -30,7 +30,9 @@ class OusterReplay : public OusterSensorNodeBase {
 
         try {
             auto meta_file = parse_parameters();
-            populate_metadata_from_json(meta_file);
+            create_metadata_publisher();
+            load_metadata_from_file(meta_file);
```
bexcite, line 34: So would user be able to replay the BAGs and omit the metadata_file path altogether if the metadata topic present in the replayed bag? 

Also how would the 2 metadata sources will be handled if one metadata is present in .bag and it's also specified as a param `metadata`? :
```
ros2 launch ouster_ros replay.launch.xml    \
    metadata:=<json file name>              \
    bag_file:=<path to rosbag file>
```

bexcite: maybe it makes sense to treat the metadata from the file as the only valid source and not make the two metadata compete? My thought is that if someone is specifying metadata para they either know that metadata topic is not present in the bag file or they want to explicitly overwrite the metadata that is recorded in the bag file with smth new.

Samahu: It does make sense, my issue with we can't selectively discard a specific topic.

The only way to achieve that is to explicitly set the topics to be replayed, which for us would be `lidar_packets` and `imu_packets`. This would in discard any other topics the user may have included in the bag (since the driver wouldn't know what is in the file).

Samahu: In a sense if the user provide metadata file we will have to exclude other topics that existed in the file. If the user doesn't then we can keep all topics.

bexcite: yeah, excluding all topics if metadata param is present maybe not as clear to the user ...  hard choices what is right, maybe we can add yet another param for this? like "exclude_other_topics" that will be "on" if the metadata is present, and if it all set in a launch file it's at least somewhat visible to the user and they can overwrite the params as they like if they want different behavior.

Samahu: I would consider this option as part of node merging effort. The reason is when this is due, raw packets won't be emitted by default by the driver. They will only be specified when the user chooses to. During bag record and replay we want to have the raw packets. I am considering to rename record.launch to record_raw.launch and replay.launch to replay_raw.launch. In this case, only we are only concerned with raw packets and we would select raw packets only along with metadata (whether as a topic or being overriden by a file). This way it is clear that this record/replay functionality is only for sensor data.

bexcite: makes sense

bexcite (approved):

PR #108, opened 2023-04-20
SW-4859: enable having multiple components of same-type under same process

### LINKED ISSUE #104: thread-caused malfunction
Hi @Samahu , there're some critical issues in ros2 version (at least ros2 version) srcs.
Case: Since it is components, I'd like to run it under container. I have two lidars and I have to run two os_cloud_node and two os_sensor_node. To obtain resulting point cloud I have to run merger_component (custom component). Now here we got a problem: to provide effective intra-process comm I have to run all five nodes under one container, so it would be one process with several threads. At this point I face at least two problems. First one is: the sensor reinits continuously. The second one is: pointcloud messages could get corrupted.
https://github.com/ouster-lidar/ouster-ros/blob/592e316c6e46ad7e1a353eb93ddf135735a119bf/ouster-ros/src/os_sensor_node.cpp#L697
https://github.com/ouster-lidar/ouster-ros/blob/592e316c6e46ad7e1a353eb93ddf135735a119bf/ouster-ros/src/os_cloud_node.cpp#L177

The problem lays in the static vars. I can't even imaging what for u've used the static var in these cases, but it is big mistake.

Samahu:
## Related Issues & PRs
- Closes #104

## Summary of Changes
- Remove the use of static vars within sensor and cloud components

## Validation
- Test running two components of the same plugin type within the same container

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -24,6 +24,8 @@ ouster_ros(2)
 * make specifying metadata file optional during record and replay modes as of package version 8.1
 * replace ``tf_prefix`` from ``os_cloud`` with ``sensor_frame``, ``lidar_frame`` and ``imu_frame``
   launch parameters.
+* bugfix: fixed an issue that prevents running multiple instance of the sensor and cloud components
+  in the same process.
 
 ouster_client
 -------------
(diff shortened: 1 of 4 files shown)


bexcite (approved):

kairenw (approved):

Aposhian:
I'm testing this, and am just getting into a loop where the init id of one of two lidars keeps changing and it just keeps reactivating it.

Aposhian:
Nevermind: it was because I was setting the same ports for both lidars

Samahu:
I did verify it to work with two sensors one container before marking the PR as ready, also have verified using two cloud nodes on a single sensor with a single container.

Aposhian:
Yes I have also verified this to work.

Issue #95, opened 2023-04-03
[ROS2] Deprecate tf_prefix and replace with sensor_frame, lidar_frame and imu_frame


doisyg:
**Describe the bug**
tf_prefix is deprecated in ROS2. We need need parameters to be able to rename the tf frame

Samahu:
@doisyg Before I proceed with your PR review, can you elaborate a bit more on why `tf_prefix` is deprecated in ROS2?

doisyg:
It was actually deprecated during the migration from tf to to tf2 in ros2, see http://wiki.ros.org/tf2/Migration#Removal_of_support_for_tf_prefix

Aposhian:
This can be closed now, right?

Samahu:
Yes, it should be closed, I thought GitHub would closed it automatically since I put "Closes" but probably I should have used "closes" word instead. I will close it manually, thanks for noting it.

Issue #33, opened 2021-09-02
Publishing pointclouds in os_sensor frame is misleading


peci1:
I know the documentation describes the situation with frames quite extensively, but generally people expect that raw data from a sensor are published in the acquisition frame. That would be `os_lidar`. But the example ROS node transforms and publishes the data in `os_sensor`, which is rotated and shifted in Z.

One possible consequence of misinterpreting the frames is usage of the scans in a raytracing pipeline. These pipelines usually take the frame_id of the incoming scan and cast rays from the origin of the given frame. But if the frame is `os_sensor`, the cast rays are invalid because the origin is shifted lower than the real optical origin of the rays.

I thus suggest changing the data frame to `os_lidar`. Users who do not care about the origin of observation should not be affected (except receiving the data in a different frame), and users interested in the observation point would get the expected data by default.

Aposhian:
Is there a reason why the frame on the messages can't be changed to the `lidar_frame`?

Samahu:
Unfortunately, I don't have enough background on what or why this was the choice but I'll check on the subject. Especially for the ROS2 driver since it is still in the stage of BETA and we need to get this right.

Aposhian:
Based on this documentation, the raw UDP data comes not as XYZ at all, but then can be transformed to the lidar frame, and then transformed to the sensor frame. It will take me some digging in the Ouster SDK to confirm whether it just converts to the lidar frame or also includes the sensor frame.
https://static.ouster.dev/sensor-docs/image_route1/image_route2/sensor_data/sensor-data.html

The pointclouds are stamped with the laser data frame in the unofficial driver: https://github.com/ros-drivers/ros2_ouster_drivers/blob/5425ed023d3c1c5293a9af90d2ae81c5df889334/ros2_ouster/src/ouster_driver.cpp#L127, and by making the clouds be stamped in the lidar frame in my fork of this version, I was able to reproduce results I get with the unofficial driver.

Aposhian:
Ok it looks like the ouster sdk (as used by the driver) converts points into the sensor frame, not the data/lidar frame. So I need to figure out why it wasn't working for me...
https://github.com/ouster-lidar/ouster_example/blob/ff46c961b690daae3e18e39999ef8651081cbc66/ouster_client/include/ouster/lidar_scan.h#L354-L355

Aposhian:
[...]

PR #112, opened 2023-04-27
Ros2 tf update

Aposhian:
## Related Issues & PRs
#105 
#95

## Summary of Changes
- configure tf frames with parameters not a single prefix
- send static transforms only once with a static transform broadcaster

## Validation
In progress


Aposhian:
Ok @Samahu I have updated this to include the latest merged PR, so now this only adds the static broadcaster

Samahu:
If static broadcast is the appropriate (for ROS2) then we don't need to add the static publisher in the C++ code.. we could just utilize the static publishers that are created when RVIZ is launched:

* In XML:
  - https://github.com/ouster-lidar/ouster-ros/blob/513c826bd59f2679575320abe1191ee208be8cd8/ouster-ros/launch/rviz.launch.xml#L21
  - https://github.com/ouster-lidar/ouster-ros/blob/513c826bd59f2679575320abe1191ee208be8cd8/ouster-ros/launch/rviz.launch.xml#L24

* In Python
  - https://github.com/ouster-lidar/ouster-ros/blob/513c826bd59f2679575320abe1191ee208be8cd8/ouster-ros/launch/rviz.launch.py#L36
  - https://github.com/ouster-lidar/ouster-ros/blob/513c826bd59f2679575320abe1191ee208be8cd8/ouster-ros/launch/rviz.launch.py#L43

Basically all we need to do is enable these static publisher regardless if RVIZ is launched or not.

Samahu:
I take that back we need the C++ code since we don't have the transform until metadata is received.

Samahu (approved):
Looks good to me.

Samahu:
@Aposhian if you don't have any other changes I am ready to merge

Aposhian:
I don't have any more changes, but I am still testing to validate that it works as expected.

Samahu:
> I don't have any more changes, but I am still testing to validate that it works as expected.

I appreciate any help!

Aposhian:
I've confirmed this works, although it gave me a little bit of unexpected results due to #33. This PR should be good to merge!

Samahu:
@Aposhian I went ahead and merged your PR into a branch so I would be able to document your changes and also disable the static publishers that I had for the rivz node.

Aposhian:
Sounds good!

Issue #55, opened 2023-01-27
Ouster (sensor.launch) Retry connecting to sensor instead of crashing


hirenpatel1207:
Hi,

Right now if we do "roslaunch sensor.launch" and if the sensor is not connected to the systems (lets say it was not powered on), then the launch file crashes.
Is there a way to retry instead of crashing? The launch file is used with other sensor (like a camera, imu etc) and crashing of this file causes the whole system to crash.

Samahu:
Hi @hirenpatel1207, there is plans to support similar things in the future but I wanted to check if you happen to try out the node attributes **required** and **respawn** and saw if they would help in your case, basically setting **required** to `false` and **respawn** to` true`.

hirenpatel1207:
Hi @Samahu ,
I tried setting `required` to `false` and `respawn` to `true` on `os_nodelet_mgr` and/or `os_node` (tried different combinations). But none of them worked. Following is my output:
```
[ INFO] [1675066044.701190001]: Initializing nodelet with 16 worker threads.
[ INFO] [1675066045.697216986]: Loading nodelet /ouster/os_node of type nodelets_os/OusterSensor to manager os_nodelet_mgr with the following remappings:
[ INFO] [1675066045.728023349]: Will use automatic UDP destination

[ INFO] [1675066046.696662483]: Loading nodelet /ouster/os_cloud_node of type nodelets_os/OusterCloud to manager os_nodelet_mgr with the following remappings:
[ INFO] [1675066046.703412536]: Loading nodelet /ouster/img_node of type nodelets_os/OusterImage to manager os_nodelet_mgr with the following remappings:
[ERROR] [1675066080.408690431]: Error setting config:  CurlClient::execute_get failed for the url: [http://os-122251000958.local/api/v1/system/firmware] with the error message: Couldn't connect to server
[FATAL] [1675066080.409086889]: Failed to load nodelet '/ouster/os_node` of type `nodelets_os/OusterSensor` to manager `os_nodelet_mgr'
[ INFO] [1675066080.414468138]: waitForService: Service [/ouster/get_metadata] has not been advertised, waiting...
[...]

PR #119, opened 2023-04-30
Handle cases of the lidar sensor becoming unpowered during operation

tombrodie:
## Related Issues & PRs
* https://github.com/ouster-lidar/ouster-ros/issues/55

## Summary of Changes
* Introduces a new launch argument `retry_configuration` (default `false`) which, when `true`, will continue to attempt to configure the lidar sensor if the sensor is uncontactable on initialisation instead of the current behaviour of the code crashing. Additionally, if the sensor becomes uncontactable during operation, this branch periodically attempts to re-establish a connection to the sensor using the configuration provided on initialisation.

## Validation
* Tested with an OS0 lidar sensor for the cases of:
1. A powered sensor on code launch
2. An unpowered sensor on code launch
3. A powered sensor on code launch, then made unpowered, then made powered again


Samahu:
included/resolved by #362

PR #123, opened 2023-05-04
change pointcloud frame to lidar_frame

Aposhian:
## Related Issues & PRs
#33

## Summary of Changes
Publish pointclouds with frame in acquisition frame.


Aposhian:
I have also tested this, and found it to be backward compatible with https://github.com/ros-drivers/ros2_ouster_drivers

Aposhian:
This PR does serve to make it backward compatible with `ros2_ouster_drivers`, but that is because of a bug in that driver. See #33

Issue #110, opened 2023-04-24
Pointcloud partially flickering (incomplete) in rviz.


batzor:
**Describe the bug**
PointCloud flickering in RVIZ. `ros2_ouster_driver` works without problem.

**To Reproduce**
Steps to reproduce the behavior (steps below are just an example):
1. ROS Foxy
2. Clone `ros2-foxy` branch and colcon build. 
3. `ros2 launch ouster_ros sensor.independent.launch.py`
4. params:
```
ouster/os_sensor:
  ros__parameters:
    sensor_hostname: '192.168.2.82'
    udp_dest: '192.168.2.81'
    mtp_dest: ''
    mtp_main: false
    lidar_mode: '512x10'
    timestamp_mode: 'TIME_FROM_ROS_TIME'
    udp_profile_lidar: 'LEGACY'
    metadata: ''
    lidar_port: 7502
    imu_port: 7503
ouster/os_cloud:
  ros__parameters:
    tf_prefix: ''
    timestamp_mode: 'TIME_FROM_ROS_TIME'
```

**Screenshots**
![Screencast 2023-04-24 15_22_16](https://user-images.githubusercontent.com/32958247/233916464-4dc68756-5d06-4688-a0a4-15aacefc999d.gif)

**Platform (please complete the following information):**
- Ouster Sensor? OS-0-64-U13
- Ouster Firmware Version? v2.4.0
- ROS version/distro? Foxy 
- Operating System? Ubuntu 20.04
- Machine Architecture? ARM
- git commit hash (if not the latest).

Samahu:
Do you connect to the sensor directly via an Ethernet cable or do you have a usb-ethernet converter in the middle?

In case you have used our ROS1 driver, have you observed similar behavior or you have only seen this issue when trying to integrate with the ROS2 driver?

I do suspect that this has to do with the ROS2 QoS settings. The driver currently uses SensorData QoS which uses "Best Effort" and has only 5 depth for the history. I do have ticket to make this configurable but if you are familiar with ROS2 and C++ you could try fiddling with them and see for example selecting "Reliable" QoS policy and/or increasing the history depth more than 5 helps.

rjwb1:
I'm also experiencing this issue on a pair of OS1-32 using a Jetson Orin with Ubuntu 20 and a 1Gbt Network switch. I have another system with this configuration that doesn't seem to have this issue. I am running ros 1 noetic.

Samahu:
[...]

PR #139, opened 2023-05-16
SW-4997: Switch from using ROS timer to threads for polling lidar data

### LINKED ISSUE #120: Specify param defaults for non-required params
If parameters are not set when they are declared without a default, in ROS 2, this crashes the node.

Any parameter that is not required to be set by the user (ex: udp_dest, mtp_main, metadata, etc.) should have a reasonable default value for the parameter.

The workaround is to set all params, and set unneeded ones to empty strings.

- Ouster Sensor? N/A
- Ouster Firmware Version? N/A
- ROS version/distro? ROS 2 Humble
- Operating System? N/A
- Machine Architecture? N/A

Samahu:
## Related Issues & PRs
- humble #140
- related #110 
- closes #120

## Summary of Changes
- Deprecate the use of timers to poll data

## Validation
- Existing functionality not affected

### DIFF
--- a/ouster-ros/src/os_sensor_node.cpp
+++ b/ouster-ros/src/os_sensor_node.cpp
@@ -174,15 +174,15 @@ class OusterSensor : public OusterSensorNodeBase {
    private:
     void declare_parameters() {
         declare_parameter("sensor_hostname");
-        declare_parameter("metadata");
-        declare_parameter("udp_dest");
-        declare_parameter("mtp_dest");
+        declare_parameter("metadata", "");
+        declare_parameter("udp_dest", "");
+        declare_parameter("mtp_dest", "");
         declare_parameter("mtp_main", false);
         declare_parameter("lidar_port", 0);
         declare_parameter("imu_port", 0);
-        declare_parameter("lidar_mode");
-        declare_parameter("timestamp_mode");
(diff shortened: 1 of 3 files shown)


ouster-ros/src/os_sensor_node.cpp
```diff
@@ -90,21 +90,8 @@ class OusterSensor : public OusterSensorNodeBase {
...
-                    if (reset_in_progress) return;
-                    auto& pf = sensor::get_format(info);
-                    connection_loop(*sensor_client, pf);
-                });
-        } else {
-            connection_loop_timer->reset();
-        }
-
-        reset_in_progress = false;
+        start_sensor_connection_thread();
```
bexcite, line 94: I presume it's not possible to reach this line couple of times and `on_activte()` can happen only once, so we don't need to worry about checking do we have any running thread already?

Samahu: We shouldn't worry about it. These are state managed .. the `on_activate` could only trigger if are at an inactivated state.

bexcite: Good good!

bexcite (approved):

Issue #144, opened 2023-05-25
corrupt scans


Janphr:
Hello!

I've been having problems lately with incomplete scans on both my OS1 (Rev 06 and D). 

I'm pretty sure I didn't have the problem in the past, but now, under some CPU load (e.g. with 'stress -c 4'), blocks of columns are missing.

Since I don't need 10hz and couldn't fix the problem by running different firmware and ouster-ros versions, I just filter out corrupt scans. These images should show the problem pretty well, with a larger and smaller area missing. These flicker over the entire 360 degrees and can be as large as 180 degrees. But since I only need the front part of the scan, I ignore the rest (the gray part) and only filter the scans that are corrupted in the front part.

![img_damaged_scan](https://github.com/ouster-lidar/ouster-ros/assets/45623010/41dee0ea-01a6-421d-9c1f-7f6ac12b3e4e)
![scan_damaged_scan](https://github.com/ouster-lidar/ouster-ros/assets/45623010/6d01d9db-8153-4829-ab91-d2a582df38a2)

- Ouster Sensor? 
OS1 Rev 06 and D
- Ouster Firmware Version? 
2.51 and 2.4
- ROS version/distro? 
noetic
- Operating System? 
linux 20.04
- Machine Architecture? 
i7-8565U (1.80GHz x 8)

Samahu:
Thanks for reporting the issue! Soon (early June) would should have an update that should reduce overall CPU utilization. I would expect it to help with your situation. I will keep you updated.

Janphr:
I forgot to mention, I have the same problem using Ouster Studio under Windows. 
Not as bad tho, but with no additional CPU load...
https://nextcloud.telematik-zentrum.de/index.php/s/rJYqg7TAPxAQsEL

With ouster-ros on linux, the flickering occurs more often and affects some larger areas. 
Sometimes several scans in a row are affected.

Issue #137, opened 2023-05-11
Frame Rate Drop


mserour:
I have an OS1-64 running using ROS2 Foxy Driver, Ubuntu 20.04 that I received yesterday. My machine is quite powerful since it has the latest processor from Intel I9-13900K and I am using the standard launch file command to run the LIDAR. The LIDAR starts to consistently publish the data at 10 FPS, however, with time evolves, the FPS keeps dropping and there is some sort of skittering/discontinuity in the LIDAR visualization from RVIZ. 

When we tried to update the resolution of the output LIDAR image to 512x10, the rate was more consistent, however, it isn't that consistent over time (observed by hz the topic). Also one more observation, if we covered the LIDAR i.e. no points, the rate is consistent since there are no points. I believe it has something to do with the processing done to output the LIDAR image.

Any recommendations?

Thank you.

mserour:
I think it is a bit different because what is happening is some point cloud frames suddenly stop and then continue to publish again. It suddenly freezes on RVIZ and then it works normal again. Therefore, the frame rate drops when trying to ros topic hz the topic it self. Also, we didn't launch the node that publishes the LIDAR image and right now RVIZ is just visualizing the point cloud but the same issue happens that the point cloud freezes from time to time.

Samahu:
> The LIDAR starts to consistently publish the data at 10 FPS, however, with time evolves, the FPS keeps dropping and there is some sort of skittering/discontinuity in the LIDAR visualization from RVIZ.

How soon it takes to observe the drop in the frame rate?

mserour:
I think it is observed immediately, some frames freeze after operation. I am sorry the word of skittering or discontinuity is misleading and incorrect. It is different issue than 110. I think the problem right now is the frames freeze as visualized in RVIZ for some reason and then the average FPS is dropped because you lost some frames.

Samahu:
I carried out few tests around this issue, here are few findings:
- There is a noticeable drop in the frame rate when using the driver with ros-foxy
- It didn't matter whether I have used **FastRTP** or **CycloneDDs**, same behavior (I haven't tested other RMW since they are slightly more involved).
- This behavior is not observed when using the rolling distro on Ubuntu 20.04 machine.. meaning I would receive 10Hz on the `/ouster/points` and the rate is consistent. Also verified it visually, i.e there is not significant drop.
- I briefly tested the driver against humble at some point and don't recall having significant frame drop but another user reported poor performance on humble #63

[...]

PR #127, opened 2023-05-09
add option for data frame

Aposhian:
## Related Issues & PRs
#33

## Summary of Changes
- add parameter to control whether points are transformed to sensor frame or left in the lidar frame, and have them stamped accordingly.

## Validation
In progress. If the TFs are setup correctly, both results should result in equivalent visualizations in rviz or foxglove.

Issue #76, opened 2023-03-09
Delay on lidar scans


SFakoorian0111:
I have a question regarding the delay on ros driver. It looks like the delay on ouster/points topic is around 120ms. Is there a way that we can reduce the delay? Also, the timestamp that we receive from the lidar scan in ros callbacks is the timestamp for the end of scan or beginning the scan. Thanks so much.
Note that we have a OS0 128 that is connected to our cpu (intel ruby R8) via Ethernet cable. The ouster setup is 512X10 in LEGACY mode, and using TIME_FROM_ROS_TIME

PR #146, opened 2023-05-31
ROS2 compatibility mode

### LINKED ISSUE #8: Homologation of ouster client for ROS / ROS 2 support
How can we combine the changes that we've made in ROS2 to enable different time sources into the ouster_client so that we can make migration between firmware version easier

### LINKED ISSUE #33: Publishing pointclouds in os_sensor frame is misleading

### LINKED ISSUE #50: Plans for ROS2?
Hi everyone,

I would like to revisit this issue [ouster-lidar/ouster_example#205](https://github.com/ouster-lidar/ouster_example/issues/205) from two years ago to get some information on you plans about ROS2.
More and more applications and drivers are shifting to ROS2 which makes it increasingly frustrating to work with ROS1 packages.

I am aware that there is a ROS2 driver by the community ([https://github.com/ros-drivers/ros2_ouster_drivers](https://github.com/ros-drivers/ros2_ouster_drivers)), but it still lacks some key features such as the Dual Return mode.

From Ouster distribution we got the information, that a official ROS2 driver is in the works and could potentially be released in the first half of 2023. Here on github it more seems like you try to cooperate with the community driver.

Could elaborate a bit on what to expect from the upcoming months as a user. It is very time consuming to implement things like ROS bridge or switching Distros, so this could help us a lot with planing on where to put resources.

Thanks in advance!

### LINKED ISSUE #63: Low performance ROS2 Humble Beta
I'm trying out the new ros2 branch just mentioned here.
#https://github.com/ouster-lidar/ouster-ros/issues/50#issuecomment-1438786002

Looking at the PointCloud in Rviz2, it is visible that the PointCloud is not updating at the 10Hz rate expected when running in the 1024x10 mode.
Running ros2 topic hz on the /ouster/points topic shows this as well.
Screenshot from 2023-02-21 12-08-23
However, the images appear to be updating at 10Hz as expected.
Screenshot from 2023-02-21 12-09-54
Screenshot from 2023-02-21 12-16-44

I'm using an OS2-64 connected directly to a NUC 12. Ubuntu 22.04 ROS Humble.
Let me know if there any anything I can try to get the PointCloud publishing at 10Hz and help out.

Samahu:
## Related Issues & PRs
- foxy's #149
- resolves #8
- resolves #50
- resolves #33
- resolves #63
- partially addresses #69
- related #76
- related #110
- incorporates #127
- related #137
- related #144

## Summary of Changes
- Build compatibility mode with [Steve Macenski's implementation of Ouster driver for ROS2](https://github.com/ros-drivers/ros2_ouster_drivers)
- Combine the functionality of **os_sensor** and **os_cloud** into a single node/component named **os_driver**
- Retain the **os_sensor** and **os_cloud** for backward compatibility and also still be able to record the raw packets
- Add unit tests around the **ThreadSafeRingBuffer**
- Add an option to select the frame of reference when publishing point cloud and use `lidar_frame` by default
- Provide topic mapping that is compatible with Steve Macenski's Ouster ROS2 driver
  - Provide support for parsing Steve's Ouster ROS2 yaml format
    - Partially support `proc_mask` handling (all flags supported ~excluding the `IMG`~)
  - Create and utilize an RVIZ config file that immediately maps to the topics of the community driver
- Add the ability to publish **sensor_msgs::msg::LaserScan**
  - Added a `scan_ring` parameter to enable picking a specific ring.
- Add ability to override QoS settings
- Factored out the functionality of OusterImage and integrate into OusterDriver same as os_cloud

### Performance Comparison
Still in the process of running a full benchmark but in general the current data shows the following:
- Improved overall performance over the current mainline driver when using the new **os_driver** node (a reduction in CPU utilization by 30%)
- Improved overall performance and stability over the community driver when using **os_driver** node (a reduction in CPU utilization by 15-20%)

## TODOs
- ~Provide topic mapping that is compatible with Steve Macenski's Ouster ROS2 driver~
- ~Provide support for parsing Steve's Ouster ROS2 yaml format~
  - ~Support the proc_mask parameter~
- ~Add an option to use select the **frame_id** for published point clouds between `os_lidar` and `os_sensor`.~
- ~Hook up the implemented **ThreadSafeRingBuffer** to minimize packets drop in case the processing thread falls behind.~
  - ~Add units around the **ThreadSafeRingBuffer**~
  - ~Reduce the number of synchronization objects needed~
- ~Add the ability to publish **sensor_msgs::msg::LaserScan**~
  - ~Configurable ring (De-scoped)~ 
- ~Configurable QoS (allow system default qos)~
- ~Factor out the functionality of OusterImage and integrate into OusterDriver same as os_cloud (De-scoped)~

### Old Interface Maintained
The three launch commands should be fully preserved under the new changes
```bash
# regular launch
ros2 launch ouster_ros sensor.launch.xml    \
    sensor_hostname:=<sensor host name>
# record
ros2 launch ouster_ros record.launch.xml    \
    sensor_hostname:=<sensor host name>     \
    bag_file:=<optional bag file name>      \
    metadata:=<json file name>              # optional
# replay
ros2 launch ouster_ros replay.launch.xml    \
    bag_file:=<path to rosbag file>         \
    metadata:=<json file name>              # optional if bag file has /metadata topic
```
### Single Node (os_sensor+os_cloud+os_image)
To use the new combined node **os_driver** run the launch file `driver.launch.py` as shown below but please find the `driver_config.yaml` and adjust to match your sensor settings:
```bash
ros2 launch ouster_ros driver.launch.py
```
### Single Node (Compatibility Mode) 
To have the same topic names as the community driver use `driver_launch.py` launch file instead, this launch file points to the `community_driver_config.yaml` and it should work seamlessly ~except for the `IMG` option of **proc_mask** parameter~ (all flags should work as expected)
```bash
ros2 launch ouster_ros driver_launch.py
```

### proc_mask flags
- When running in a single node test enabling \ disabling individual flags for the `proc_mask`.
  - For example enable only the `IMU`, `SCAN`, `IMG` and `PCL`.
  - Watch that only the specific message of the enabled message type actually gets published

### Alternate QoS settings
- Test enabling \ disabling the `use_system_default_qos`
Note that when setting `use_system_default_qos` to **true** RViz won't be able to display the point cloud by default. You would need to select `/ouster/points` topic and adjust the policy to be reliable.

### Alternate PointCloud frame
- Test altering `point_cloud_frame` between `os_lidar` and `os_sensor`
Altering between the two should be clearly reflected in the the rendered point cloud in RVIZ

### Sensor Lifecycle Management
- Connect the sensor, then invoke the `/ouster/reset` service
  ```bash
  ros2 service call /ouster/reset std_srvs/srv/Empty
  ```
  - Watch point cloud update pauses for couple of second while the sensor restarts then the node resumes normally. 
- Connect to the sensor, open the sensor configuration change and apply any change except the udp_lidar_profile. 
  - Watch the os_driver node detects the change and re-adapts accordingly.
- Connect to the sensor, open the sensor configuration change and change the udp_lidar_profile. 
  - Watch the sensor, generates few errors and then performs successfully re-activates itself.


ouster-ros/src/lidar_packet_handler.h
```diff
@@ -0,0 +1,288 @@
...
+        uint64_t scan_ns = curr_scan_first_nonzero_idx == 0
+                               ? curr_scan_first_nonzero_value
+                               : impute_value(last_scan_last_nonzero_idx,
+                                              last_scan_last_nonzero_value,
+                                              curr_scan_first_nonzero_idx,
+                                              curr_scan_first_nonzero_value,
+                                              static_cast<int>(ts_v.size()));
+
+        last_scan_last_nonzero_idx =
+            find_if_reverse(ts_v, [](uint64_t h) { return h != 0; });
```
celentes, line 214: [nit]
works the same way as
```
std::find_if(std::make_reverse_iterator(ts_v.data() + ts_v.size()),
             std::make_reverse_iterator(ts_v.data()),
             [](auto&&){ .... });
```

ouster-ros/include/ouster_ros/os_processing_node_base.h
```diff
@@ -17,21 +17,14 @@ namespace ouster_ros {
 
 class OusterProcessingNodeBase : public rclcpp::Node {
    protected:
-    explicit OusterProcessingNodeBase(const std::string& name,
+    OusterProcessingNodeBase(const std::string& name,
                                       const rclcpp::NodeOptions& options)
```
celentes, line 21: [nit] formatting

ouster-ros/src/os_sensor_node.h
```diff
@@ -0,0 +1,190 @@
...
+    LifecycleNodeInterface::CallbackReturn on_configure(
+        const rclcpp_lifecycle::State& state);
+    LifecycleNodeInterface::CallbackReturn on_activate(
+        const rclcpp_lifecycle::State& state);
+    LifecycleNodeInterface::CallbackReturn on_error(
+        const rclcpp_lifecycle::State& state);
+    LifecycleNodeInterface::CallbackReturn on_deactivate(
+        const rclcpp_lifecycle::State& state);
+    LifecycleNodeInterface::CallbackReturn on_cleanup(
+        const rclcpp_lifecycle::State&);
```
celentes, line 54: [nit] missing argument name, not in line with other methods

ouster-ros/src/thread_safe_ring_buffer.h
```diff
@@ -0,0 +1,129 @@
...
+        }
+        emptyCondition.notify_one();
+    }
+
+    /**
+     * Gives access to read the buffer through a callback, the method will block
+     * until there is something to read is available.
+     */
+    template <typename BufferReadT>
+    void read(BufferReadT&& buffer_read) {
```
celentes, line 93: BufferReadT is taken as universal reference vs getting copied as BufferWriteT above.
In some edge scenarios these may well be different, it's a good idea to keep these consistent unless we have a good reason to.

celentes: On that note, do we have a convention for templated functors, like `BufferReadFn` instead of `BufferReadT`?

ouster-ros/src/lidar_packet_handler.h
Samahu: This so valid and it was what i initially attempted but it did fail to compile when using gcc-9.

celentes: _Interesting_. Okay.

ouster-ros/src/thread_safe_ring_buffer.h
Samahu: Thanks for noting that, my intention was to accept as a universal reference to avoid copying the lambda functions when possible but I don't think I am currently making use of it. I will change update all the definitions to use &&.

I don't believe we have a convention for naming template typenames but using `BufferReadFn` seems appropriate.

celentes (approved):

twslankard (approved):

ouster-ros/src/thread_safe_ring_buffer.h
```diff
@@ -0,0 +1,129 @@
...
+        return active_items_count == 0;
+    }
+
+    /**
+     * Checks if the ring buffer is full.
+     */
+    bool full() const {
+        std::lock_guard<std::mutex> lock(mutex);
+        return active_items_count == max_items_count;
+    }
```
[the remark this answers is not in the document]
Samahu: There are uses cases to these methods beyond the unit tests. For example, say I want to monitor the average utilization of the buffer from another thread. I would want to call the `size` method but wouldn't necessarily need to perform a read or write. I think adding a remark to the method description similar to python's `Queue.isEmpty` makes sense.

ouster-ros/launch/driver.launch.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def generate_launch_description():
+    """
+    Generate launch description for running ouster_ros components separately each
+    component will run in a separate process).
+    """
+    ouster_ros_pkg_dir = get_package_share_directory('ouster_ros')
+    default_params_file = \
+        Path(ouster_ros_pkg_dir) / 'config' / 'parameters.yaml'
```
kairenw, line 28: dirver_params.yaml ?

ouster-ros/config/driver_params.yaml
```diff
@@ -0,0 +1,25 @@
+ouster/os_driver:
```
kairenw, line 1: possibly im not looking in the right place. 

think it would help to specify what the behavior is if you don't fill these out / leave as is, so for ex sensor_hostname is necessary but the rest seem to be optional

kairenw (approved):

ouster-ros/launch/sensor.independent.launch.xml
```diff
@@ -44,6 +44,17 @@
   <arg name="lidar_frame" default="os_lidar" description="value can not be empty"/>
   <arg name="imu_frame" default="os_imu" description="value can not be empty"/>
 
+  <arg name="use_system_default_qos" default="false" description="Use the default system QoS settings"/>
+  <arg name="proc_mask" default="IMG|PCL|IMU|SCAN" description="
+    currently IMG is not supported and does not affect anything,
```
kairenw, line 49: specifically for this launch?

ouster-ros/launch/driver.launch.py
Samahu: Yes, this should be `driver_params.yaml`, I will change it.

ouster-ros/config/driver_params.yaml
Samahu: makes sense. I will throw few comments.

ouster-ros/launch/sensor.independent.launch.xml
Samahu: Yes, specifically for this launch file. `sensor.independent.launch.xml` still use the `os_sensor`, `os_cloud`, and `os_image`. The `os_cloud` can only control `imu`, `cloud` and `scan` messages. It can not control image messages, thus to remove image messages you would have to comment out the `os_image` node. The other launch file `sensor.composite.launch.xml` replaces the composition of 3 components with `os_driver` which supports the 4 flags.

Samahu: Yes, specifically for this launch file. `sensor.independent.launch.xml` still use the `os_sensor`, `os_cloud`, and `os_image`. The `os_cloud` can only control `imu`, `cloud` and `scan` messages. It can not control image messages, thus to remove image messages you would have to comment out the `os_image` node. The other launch file `sensor.composite.launch.xml` replaces the composition of 3 components with `os_driver` which supports the 4 flags.

PR #153, opened 2023-06-20
Add parameter to os_cloud_node to control the sensor frames publishing

abAgriRobot:
## Related Issues & PRs
- resolves: #197

## Summary of Changes
By default the **os_cloud_node** publishes the static transformation for the **sensor_frame**, **lidar_frame** and **imu_frame**.
In some cases this transformations are defined in a higher level URDF file, and having the node publishing the frames may create conflicts in the tf tree.
We added a parameter to control if the node publishes the frames. 
```
    ros__parameters:
        sensor_frame: 'os_sensor'
        lidar_frame: 'os_lidar'
        imu_frame: 'os_imu'
        timestamp_mode: "TIME_FROM_ROS_TIME" # this value needs to match os_sensor/timestamp_mode
        publish_sensor_frames: True
``` 
The parameter defaults to True so the behavior of the node doen't change unless explicitly define
```
       publish_sensor_frames: False
```

### DIFF
--- a/ouster-ros/config/parameters.yaml
+++ b/ouster-ros/config/parameters.yaml
@@ -1,18 +1,19 @@
 ouster/os_sensor:
   ros__parameters:
-    sensor_hostname: ''
-    udp_dest: ''
-    mtp_dest: ''
+    sensor_hostname: ""
+    udp_dest: ""
+    mtp_dest: ""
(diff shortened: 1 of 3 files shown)


Samahu:
Fulfilled by #393

PR #66, opened 2023-02-23
Destaggering of the point cloud

66lukas66:
## Related Issues & PRs
Issus #61 and #60 Destaggering of a Point cloud
## Summary of Changes
Adding the function destagger in ous.ros with a option to destagger or not.
## Validation
using of the new function checkofDestagger who check if a point cloud is destaggered

Issue #60, opened 2023-02-16
Destaggerd Point Cloud 


66lukas66:
I want the Ouster to have a destaggered (organized) Point Cloud published. As per the documentation located at https://static.ouster.dev/sdk-docs/reference/lidar-scan.html#staggering-and-destaggering, there is a function available for destaggering. However, I am unsure of where to implement it within the Ouster in order to obtain a destaggered point cloud.

My assumption is that the destaggering function needs to be called from within the scan_to_cloud_f function of the os_ros. But I am uncertain of the exact location within that function where the destaggering function should be integrated.
My guess is that it needs to be called inside the scan_to_cloud_f function of os_ros. But where exactly ?

Samahu:
So currently the destaggering step is only applied to the produced images, the code block that performs that can be found here:
https://github.com/ouster-lidar/ouster-ros/blob/22acf88f2f162cff62e10318aed7c51497ebc08a/src/os_image_nodelet.cpp#L137-L154
This code block is not using the destagger method since it would be much faster to do it this way when you are destaggering multiple images at the same time. rather than invoking the destagger method on each image.

So in essence, you can take the output of this method and convert it again to a PCL point cloud object and publish it on a separate topic.

On the other hand you can apply the same action or call it the destagger method before producing the first point cloud (i.e. modifying the results of the `/points` topic). If you choose this route then you should apply the destagger step on the range image before passing it to the `ouster::cartesianT` method on this line:
https://github.com/ouster-lidar/ouster-ros/blob/22acf88f2f162cff62e10318aed7c51497ebc08a/src/os_ros.cpp#L230

So you destagger the `range` image and pass the destaggered result to the `ouster::cartesian` method. It is worth noting that this will, however, affect the results of the `os_image_nodelet` since it will apply destaggering again and thus the image(s) will appear skewed if you try to visualize them in RVIZ.

Hope this answers your question

66lukas66:
Thanks for the answer but if in ouster-ros/src/os_ros.cpp before the range destagger, what parameter should pixel_shift_by_row be set to in the destagger method?

Samahu:
That parameter should always be set to `info.format.pixel_shift_by_row`, where **info** is the object you get from parsing the sensor metadata.

PR #188, opened 2023-08-09
SW-5396: publish point cloud in destaggered form

### LINKED ISSUE #61: Referece to #Destaggerd Point Cloud using of the destaggered function for Point Clouds
I want the Ouster to have a destaggered (organized) Point Cloud published. As per the documentation located at https://static.ouster.dev/sdk-docs/reference/lidar-scan.html#staggering-and-destaggering, there is a function available for destaggering. 
I know that the destagger function must be used in os_ros.cpp before line 230.

But it still doesn't work. I destaggered the range before **line 230** in **ouster-ros/src/os_ros.cpp.**
```
Eigen::Ref<const ouster::img_t<uint32_t>> refrange = range;
range = ouster::destagger(refrange,pixel_shift_by_row);
```

But the published point cloud is still staggered. Is it because a staggered range is a timestep X scanline matrix and a destaggered range is an azimuth angle X scanline matrix. However, in the function **destagger(const Eigen::Ref<const img_t<T>>& img, const std::vector<int>& pixel_shift_by_row, bool inverse)** the size of the return matrix is ​​set to the size of **img**. 
As I understand it the return matrix is a matrix with the size timestep X scanline and not a destaggerd matrix with the size of azimuth angle X scanline matrix .
 Or is the Ouster azimuth angle number equal to the number of timesteps per scan? 
And what is the reason that only the range must be destaggerd and not also the reflectivity?

Samahu:
## Related Issues & PRs
- ported from #187
- related #60 
- closes #61 
- related #66 

## Summary of Changes
- Publish de-stagggered point cloud
- wrap all classes within the **ouster_ros** namespace

## Validation
- Review test procedures

### DIFF
--- a/src/os_ros.cpp
+++ b/src/os_ros.cpp
@@ -198,13 +198,12 @@ void copy_scan_to_cloud(ouster_ros::Cloud& cloud, const ouster::LidarScan& ls,
 
 template <typename PointT, typename RangeT, typename ReflectivityT,
           typename NearIrT, typename SignalT>
-void copy_scan_to_cloud_destaggered(ouster_ros::Cloud& cloud, const ouster::LidarScan& ls,
-                        uint64_t scan_ts, const PointT& points,
-                        const ouster::img_t<RangeT>& range,
-                        const ouster::img_t<ReflectivityT>& reflectivity,
-                        const ouster::img_t<NearIrT>& near_ir,
-                        const ouster::img_t<SignalT>& signal,
-                        const std::vector<int>& pixel_shift_by_row) {
+void copy_scan_to_cloud_destaggered(
+    ouster_ros::Cloud& cloud, const ouster::LidarScan& ls, uint64_t scan_ts,
+    const PointT& points, const ouster::img_t<RangeT>& range,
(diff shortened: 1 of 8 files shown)


src/os_ros.cpp
```diff
@@ -196,6 +196,46 @@ void copy_scan_to_cloud(ouster_ros::Cloud& cloud, const ouster::LidarScan& ls,
...
+        for (auto v = 0; v < ls.w; v++) {
+            const auto col_ts = timestamp[v];
+            const auto ts = col_ts > scan_ts ? col_ts - scan_ts : 0UL;
+            const auto src_idx = u * ls.w + (v + ls.w - pixel_shift_by_row[u]) % ls.w;
+            const auto tgt_idx = u * ls.w + v;
+            const auto xyz = points.row(src_idx);
+            cloud.points[tgt_idx] = ouster_ros::Point{
+                {static_cast<float>(xyz(0)), static_cast<float>(xyz(1)),
+                 static_cast<float>(xyz(2)), 1.0f},
+                static_cast<float>(sg[src_idx]),
```
bexcite, line 228: nit: formatting

bexcite (approved):

PR #189, opened 2023-08-10
Add tf_prefix to point_cloud_frame

louis-wiesmann:
## Summary of Changes
Add the tf_prefix also to the point cloud frame. Without these changes the frame_id of the published point cloud are without the namespace prefix. Therefore, do not match with the published tf_tree where the frames do have the namespace prefix.  

## Validation
Set tf_prefix: e.g. tf_prefix="test"
check frame_id of the point cloud. 
check frames of the published tf_tree (e.g using rqt)


Samahu (approved):
on point :heavy_check_mark:

PR #193, opened 2023-08-16
disable intraprocess for transient local topics

Aposhian:
## Related Issues & PRs
N/A

## Summary of Changes
Explicitly disable intraprocess communication for metadata publisher and subscription, otherwise it fails if you try to set `use_intraprocess_comms: True` on the whole node.

## Validation
In progress.

### DIFF
--- a/ouster-ros/src/os_processing_node_base.cpp
+++ b/ouster-ros/src/os_processing_node_base.cpp
@@ -16,8 +16,10 @@ void OusterProcessingNodeBase::create_metadata_subscriber(
     auto latching_qos = rclcpp::QoS(rclcpp::KeepLast(1));
     latching_qos.reliability(RMW_QOS_POLICY_RELIABILITY_RELIABLE);
     latching_qos.durability(RMW_QOS_POLICY_DURABILITY_TRANSIENT_LOCAL);
+    rclcpp::SubscriptionOptions subscription_options;
+    subscription_options.use_intra_process_comm = rclcpp::IntraProcessSetting::Disable;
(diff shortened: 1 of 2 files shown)


Samahu:
@Aposhian is this ready? Do you plan on providing validation steps?

Aposhian:
I have tested this and it works. I will provide a test to replicate shortly

Aposhian:
Ok here are the validation steps to reproduce the bug and how this PR fixes it:

To reproduce:
- Edit `sensor.composite.launch.py` to have intraprocess comms enabled. For example on this branch: https://github.com/fireflyautomatix/ouster-ros/blob/88614044af6f0bd215b1baabee8841578d05ec2d/ouster-ros/launch/sensor.composite.launch.py#L45
- `ros2 launch ouster_ros sensor.composite.launch.py`

You get this error:
```
[component_container_mt-2] [ERROR] [1692982283.515302684] [ouster.os_container]: Component constructor threw an exception: intraprocess communication allowed only with volatile durability
```

But after applying this fix by merging this PR in like on this branch https://github.com/fireflyautomatix/ouster-ros/tree/example/intraprocess-comms-fixed, the error doesn't show up anymore.

Samahu:
@Aposhian Which rosdistro do you test against? I couldn't generate the problem under rolling.

Aposhian:
Humble. I believe there is addressed in rolling.

Samahu:
:thinking: then I need to create a branch for humble similar to having a foxy branch and merge the pull request into it. Let me get that ready and will get back to you.

PR #202, opened 2023-08-30
Fix: Make libjsoncpp-dev exec_depend match build_depend and system name

tijsvdsmagt:
There is a small issue with the current package.xml on the `ros2` branch.

`libjsoncpp` does not exist when installing with `apt`. Instead, it should be `libjsoncpp-dev`.
The current situation can cause errors for certain build systems using the `exec_depend` keys.
This PR brings it in line with the `build_depend` key for `libjsoncpp-dev`.

## Related Issues & PRs
None.
I can create one if required.

## Summary of Changes
* Appended `-dev` to `<exec_depend>libjsoncpp`

## Validation
- IMO not required as it has no effect on default builds.

### DIFF
--- a/ouster-ros/package.xml
+++ b/ouster-ros/package.xml
@@ -33,7 +33,7 @@
   <exec_depend>rosidl_default_runtime</exec_depend>
   <exec_depend>launch</exec_depend>
   <exec_depend>launch_ros</exec_depend>
-  <exec_depend>libjsoncpp</exec_depend>
+  <exec_depend>libjsoncpp-dev</exec_depend>
   <exec_depend>curl</exec_depend>
   <exec_depend>spdlog</exec_depend>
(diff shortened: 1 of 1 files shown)


Samahu:
@tijsvdsmagt which platform do you face this issue? sometimes libjsoncpp has other names on different platform but libjsoncpp  is the accepted entry name in ROS package index. So if you are installing the package through **rosdep** there shouldn't be any issue.

tijsvdsmagt:
@Samahu 

I could not find it initially, but I see that indeed for rosdep it points to `libjsoncpp1` for Debian based system. I am using a system alternative to rosdep which did not have this mapping and thus got confused because it could not find the name.

I will add the mapping in our system, so this fix would not be required indeed. Thank you for responding.

Issue #97, opened 2023-04-03
[ROS2] configurable pointcloud type


doisyg:
**Is your feature request related to a problem? Please describe.**
Generated pointcloud is too heavy with 9 fields for most use case:
- x
- y
- z
- intensity
- t
- reflectivity
- ring
- ambient
- range

> POINT_CLOUD_REGISTER_POINT_STRUCT(ouster_ros::Point,
>     (float, x, x)
>     (float, y, y)
>     (float, z, z)
>     (float, intensity, intensity)
>     // use std::uint32_t to avoid conflicting with pcl::uint32_t
>     (std::uint32_t, t, t)
>     (std::uint16_t, reflectivity, reflectivity)
>     (std::uint16_t, ring, ring)
>     (std::uint16_t, ambient, ambient)
>     (std::uint32_t, range, range)
> )

**Describe the solution you'd like**
Configurable pointcloud fields, or at least the option the choose the light PointXYZI format

dambrosio:
@doisyg that is a great idea. I also think one thing that could dramatically reduce the size of the output `PointCloud2` message would be to remove the unused/padding bytes in the `PointCloud2` `data` array.

For example, in a single `point_step` there are 18 bytes of padding:
- 4 bytes between the `z` and `intensity` fields
- 2 bytes between the `ambient` and `range` fields
- 12 bytes after the `range` field
-
This would change each `point_step` to 30 instead of 48.

For simple math, a PointCloud with 100,000 points would result in a `data` array size difference of ~46.15%. This would be 4.57MB vs 2.86MB per PointCloud2 message. 

Maybe there is a reason for padding each `point_step` but it does increase the size of each PointCloud message a thus increases the cost of processing and transmitting the messages.

dambrosio:
Also, this looks to be a duplicate of #29, may want to combine.

PR #216, opened 2023-09-13
SW-5466: Support Velodyne and other point types in ouster-ros driver

### LINKED ISSUE #15: Change the format into velodyne format
Hi everyone:

I am sorry that I am fresh and I would like to ask that how can I change the format of ouster into Velodyne like format. I found the code here:
<html>
<body>

void ouster_driver::OS1::convert2XYZIR | ( | const CloudOS1 & | in,
  |   | CloudOS1XYZIR & | out
  | )

but I did not know which code and which part of the code should I change, thx!

<br class="Apple-interchange-newline">
</body>
</html>void ouster_driver::OS1::convert2XYZIR	(	const [CloudOS1](http://docs.ros.org/en/melodic/api/ouster_driver/html/namespaceouster__driver_1_1OS1.html#a9455ea14ce69167cfdc9a34150a1fa31) & 	in,
[CloudOS1XYZIR](http://docs.ros.org/en/melodic/api/ouster_driver/html/namespaceouster__driver_1_1OS1.html#ac346ac2772c984d044b5d6c3d808b3fa) & 	out 
)

### LINKED ISSUE #29: Parameter for Controlling PCL PointType
Hello,

Thanks for the Ouster driver, it's works well! I would like to make one request, which is to my knowlesge, the user does not have any control over what fields the Ouster publishes. The driver will always publish X, Y, Z, intensity, t, reflectivity, ring, noise, and range. This is a lot of data that to most users is not necessary. In my specific application, I want XYZI because I am fusing the point cloud with XYZI point clouds from other depth sensors onboard a robot. Therefore, having the same PointType is convenient. I can envision other folks wanting to disable fields they don't need to reduce the amount of computational/communications resources required to run the driver.

In order to resolve this issue, I had to make changes to the driver source code (ouster_ros/include/ouster_ros/point_os1.h), which I would prefer to not do.

Hacky solution: commented out unwanted PCL fields

Thanks!

Gene

### LINKED ISSUE #210: Add low data rate point type
I would like the pointclouds to use less bandwidth when I use the [low data rate profile](https://static.ouster.dev/sensor-docs/image_route1/image_route2/sensor_data/sensor-data.html#low-data-rate-profile) from the sensor.

When the low data rate profile is selected, use a different point type that has lower precision and fewer fields. For example:

```cpp
struct EIGEN_ALIGN16 LowDataRatePoint {
    PCL_ADD_POINT4D;
    uint32_t t;
    uint8_t reflectivity;
    uint16_t ring;
    uint8_t ambient;
    uint16_t range;
    EIGEN_MAKE_ALIGNED_OPERATOR_NEW
};
```

Not sure.

- Ouster Sensor? Rev 6 and Rev 7
- Ouster Firmware Version? 2.3+
- ROS version/distro? N/A
- Operating System? N/A
- Machine Architecture? N/A

Samahu:
## Related Issues & PRs
- resolves #15 
- resolves #29
- resolves #210 
- related #97
- related #217
- related #223

## Summary of Changes
- added the ability to customize the published point clouds(s) to velodyne point cloud format and
  other common pcl point types.
- made ouster_image_component operate separately from ouster_cloud_component

## Validation
- Run the driver with default parameters and verify no observable behavioral changes.
- Change `point_type` parameter to `native` and verify that it reflects the same point cloud format that we get from the sensor.
- Change `point_type` parameter to `xyz` and verify that the published point cloud format only has `xyz` fields, regardless of the underlying sensor format.
- Change `point_type` parameter to `xyzi` and verify that the published point cloud format only has `xyz` fields plus intensity, regardless of the underlying sensor format (you should get a warning when the low data mode is selected for this mode).
- Change `point_type` parameter to `xyzir` and verify that the published point cloud format only has `xyz` fields plus intensity plus ring, regardless of the underlying sensor format (you should get a warning when the low data mode is selected for this mode).


ouster-ros/config/driver_params.yaml
```diff
@@ -67,3 +67,18 @@ ouster/os_driver:
...
+     # point_type[optional]: choose from: {original, native, xyz, xyzi, xyzir}
+     # Here is a breif description of each option:
+     #  - original: ouster_ros::Point
+     #  - native: directly maps all fields as published by the sensor to an
+     #          equivalent point cloud representation with the additon of ring
+     #          and timestamp fields.
+     #  - xyz: the simplest point type, only has {x, y, z}
+     #  - xyzi: same as xyz point type but adds intensity(signal) field. this
+     #          type is not compatible with the low data profile.
+     #  - xyzi: same as xyzi type but adds ring(channel) field.
```
mtswisher, line 79: xyzi should be xyzir on this line

ouster-ros/include/ouster_ros/sensor_point_types.h
```diff
@@ -0,0 +1,336 @@
...
+ *
+ * @file sensor_point_types.h
+ * @brief native sensor point types
+ * The following are one-to-one mapping of pcl point representatios that could
+ * fit the data sent by the sensor (with the addition of t and ring fields),
+ * All of these representation follow the same order of LaserScan fields:
+ * 1. range
+ * 2. signal
+ * 3. refelctivity
+ * 4. nea_ir
```
mtswisher, line 13: near_ir :D

ouster-ros/include/ouster_ros/sensor_point_types.h
```diff
@@ -0,0 +1,336 @@
...
+ * The only Point_RNG15_RFL8_NIR8 aks LOW_DATA which does not omits the signal
```
mtswisher, line 14: aka as opposed to aks

mtswisher (changes requested):
Just made some typo changes :)  How do you decide whether to designate a function as inline? I haven't seen the factory pattern in quite awhile so it was cool to see that solve this problem. Will probably need to dig into some more to really understand what is going on but LGTM :)

Samahu:
> Just made some typo changes :)

Addressed!

> How do you decide whether to designate a function as inline?

Typically for short one liner methods this could minimally increase the code size but avoid having to invoke the method which is always faster .. With modern compilers the **inline** keyword only serves as a suggestion, compilers could inline methods automatically .. and could also ignore the suggestion if the method body is too long (in this case one needs to use **__forceinline**).  
 
> I haven't seen the factory pattern in quite awhile so it was cool to see that solve this problem. Will probably need to dig into some more to really understand what is going on but LGTM :)

Yeah, the factory pattern is definitely needed here. I will add more comments to the PR to explain what's going on and how could developers use existing code to implement other point types that are not included here.

twslankard (approved):

Issue #190, opened 2023-08-11
ROS2 driver pointcloud data lost


ahuazuipiaoliang:
**Describe the bug**
sometimes the pointcloud data lost when use the ROS2 driver for OS1 128 REV7

**To Reproduce**
Steps to reproduce the behavior (steps below are just an example):
1. source ros environment
2. compile the project workspace
3. source the project workspace
4. ros launch ouster_ros sensor.launch.xml
5. automatically open Rviz2
6. observe the issue ...

**Screenshots**
![image](https://github.com/ouster-lidar/ouster-ros/assets/35034146/c9d7e961-6287-4e0a-877f-6671337b6ed9)
abnormal data
![image](https://github.com/ouster-lidar/ouster-ros/assets/35034146/50bf4d19-694e-4e91-8b66-86f6b332bd22)
normal data

**Platform (please complete the following information):**
- Ouster Sensor? 
OS1 128 REV7
- Ouster Firmware Version? 
![image](https://github.com/ouster-lidar/ouster-ros/assets/35034146/662aa504-22e4-4454-ba04-77125244e920)
- ROS version/distro? 
Humble
- Operating System?
Ubuntu 22.04
- Machine Architecture? 
x86
- git commit hash (if not the latest).
13171a84b986cae489f644e86fb4a57a4b373cf9

ahuazuipiaoliang:
more abnormal screenshots
![image](https://github.com/ouster-lidar/ouster-ros/assets/35034146/2d490618-a518-4a4c-adfe-43fde8fd6087)
![image](https://github.com/ouster-lidar/ouster-ros/assets/35034146/3ae0305d-d8a6-4dc2-9563-5e98eee22165)
![image](https://github.com/ouster-lidar/ouster-ros/assets/35034146/b20e5978-d624-47d9-8758-4c3ae7487b69)

Samahu:
Hi @ahuazuipiaoliang, we do have have some upcoming changes the would help with the issue. In the meantime, could you try to increase the receive buffer  on our system and see if that helps out:
```bash
sudo sysctl -w net.core.rmem_default=16777216
sudo sysctl -w net.core.rmem_max=16777216
```

ahuazuipiaoliang:
@Samahu I found the data lost problem was caused by the hardware, since I switched to a new LiDAR the problem disappeared. I will close the issue.

Samahu:
Fixed per #262

PR #229, opened 2023-10-12
Configurable qos and default history depth

Aposhian:
## Related Issues & PRs
#190
https://github.com/ouster-lidar/ouster_example/issues/120

## Summary of Changes
Set history depth on lidar packets topic to be something more reasonable.

Allow for configuring qos of the packet topics using parameters. http://design.ros2.org/articles/qos_configurability.html
Note that this requires at least ROS Galactic. It is not compatible with Foxy or earlier. If that backward compatibility is desirable I would like to instead expose a parameter that at least can set the history depth.

This is desirable because neither the system default qos nor the sensor data qos profiles have very large history depths, and the lidar packet topic can be publishing at hundreds of hertz, and I believe it might be dropped packets that are causing missing data in the pointcloud. For example:
```yaml
os_sensor:
  ros__parameters:
    qos_overrides:
      /lidar_packets:
        publisher:
          depth: 1000
os_cloud:
  ros__parameters:
    qos_overrides:
      /lidar_packets:
        subscription:
          depth: 1000
```

Note that this also applies to intraprocess communication since the intraprocess buffer is set to the size of the history depth qos.

## Validation
Lost packets show up if the os_cloud process is under high load, resulting in a "propeller effect". I can confirm that increasing the history depth eliminates this issue.

### DIFF
--- a/ouster-ros/src/os_cloud_node.cpp
+++ b/ouster-ros/src/os_cloud_node.cpp
@@ -157,7 +157,7 @@ class OusterCloud : public OusterProcessingNodeBase {
             rclcpp::SubscriptionOptions subscription_options;
             subscription_options.qos_overriding_options = rclcpp::QosOverridingOptions::with_default_policies();
             lidar_packet_sub = create_subscription<PacketMsg>(
-                "lidar_packets", selected_qos,
+                "lidar_packets", selected_qos.keep_last(100),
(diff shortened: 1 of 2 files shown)


nfa-ahrnbom:
It's been years, but this is still very relevant :laughing:

PR #233, opened 2023-10-16
Reconnection behavior

lowellausen:
## Related Issues & PRs

This same reconnection behavior for ROS1: https://github.com/ouster-lidar/ouster-ros/pull/119
The issue it started from: https://github.com/ouster-lidar/ouster-ros/issues/55

## Summary of Changes

Idea adapted to ROS2 from this PR: https://github.com/ouster-lidar/ouster-ros/pull/119 

Added reconnection behavior

- New `retry_configuration` parameter that sets if we want to use the new reconnection behavior or not. Initially I've defaulted it to `false` as not to change current behavior automatically, ideas?
- Makes config an attribute so we can reuse it in the connection loop
- Adds logic to the `configure_sensor()` and to `connection_loop()` methods so that we can retry to configure the sensor
- [QUESTIONABLE CHANGE MARK] Added throttling to the logs that come when trying to reconnect, to avoid the consequent spamming

## Validation

- Starting the system with sensor and/or ethernet cable disconnected
  - If `retry_configuration`
    - Driver will keep trying to configure the sensor until it's connected
    - Logs are throttled 
  - Else
    - Same behavior as before: driver will fail to configure the sensor and stop trying
- Same if the sensor is disconnected during runtime after it's already been configured

### DIFF
--- a/ouster-ros/config/driver_params.yaml
+++ b/ouster-ros/config/driver_params.yaml
@@ -67,3 +67,4 @@ ouster/os_driver:
     # data QoS. This is preferrable for production but default QoS is needed for
     # rosbag. See: https://github.com/ros2/rosbag2/issues/125
     use_system_default_qos: false
+    retry_configuration: false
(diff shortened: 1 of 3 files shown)


Samahu:
included/resolved by #362

Issue #10, opened 2021-09-15
Loss lidar packet messages (ROS melodic)


rdouguet:
Hello,

I'm trying to record data provided by a lidar OS1-64 and by a basler insdustrial camera on a Jetson Xavier board.  For that, I use the two ros packages ouster_ros and pylon_camera. These two sensors are connected via Ethernet and each camera uses a specific IP address (the thoughput on the GiE is not saturated ~ 50%).

When I use OS1-64 alone, I get :
- /os_node/lidar_packet	**76650 msgs** by minute
- /os_node/lidar_imu	5988 msgs by minute

But when I use the two sensors together, all is ok for the camera but the messages falling down to **366** for the /os_node/lidar_packet... There are clairly a conflict between this two packages and ethernet configuration but I don't understand why ?  Especially since with another lidar and package (hesai), I don't have this problem.

I seen similary issue in the history github "Messages get dropped [ros melodic + industrial camera] ouster-lidar/ouster_example#225" but there is no solution.

Thanks for your help.

ShepelIlya:
I am updated my sensors to 2.4 fw and update my drivers version to commit e4bd020. In my version of ouster driver i have little upgrades for multicast and cloud filtering.

My previous setup was on fw 2.3 with my custom code based on this commit of [ouster_example](https://github.com/ouster-lidar/ouster_example/commit/43107a1d14b8cd22378a830bd07aa7acca1269a4).

On old setup (3 drivers for 2 os1-32 and 1 os0-64 sensors running on jetson AGX xavier) htop showed a total CPU load of 17-20 percent.
On new setup (2 drivers for 2 os1-32 on same AGX xavier) total CPU load is 70-80 percent.

Why CPU usage is so high on new drivers and how we can fix it?
[Is it ok to use ros::Timer for connection_loop? (and what about ros::Duration(0)?)](https://github.com/ouster-lidar/ouster-ros/blob/master/src/os_sensor_nodelet.cpp#L339-L341)

I will try to find the reason for this behavior, but if you have any idea what the problem is, then tell me about it please.

Thanks for your help.

Samahu:
[...]

PR #249, opened 2023-11-03
SW-5622: Bump up ouster_client to 20231031 release

### LINKED ISSUE #247: Increase RCVBUF_SIZE to prevent package loss
We experience package losses when it comes to some system load on our vanilla Ubuntu 20.04.
We reproducibly tested it with maxing out the system's load with `$ yes` while running:
```
roslaunch ouster_lidar sensor.launch sensor_hostname:=$OUSTER_OS1_SENSOR_HOSTNAME udp_dest:=$OUSTER_OS1_UDP_DESTINATION lidar_port:=$OUSTER_OS1_UDP_PORT_LIDAR 
``` 
The applications were compiled with `-O3` and `release` flags beforehand.
However, adapting the system's `rmem_default` and `rmem_max` to e.g., 20MB did not help, since the ouster_client sets the receive-buffer explicitly to 256kB in the [client.cpp](https://github.com/ouster-lidar/ouster_example/blob/d73079865ebadd961d7af81b235bb34747f7cb6a/ouster_client/src/client.cpp#L56)

Increasing `RCVBUF_SIZE` to 20MB as well (this value was just a random guess BTW!) solved all our problems and we don't experience any frame drops from now on.

However, I think that just increasing `RCVBUF_SIZE` is a bad practice.
I think, that this value should be exposed as an argument such that it can be configured concerning the user's demand.

**What do you think?** 

Example with package loss:
image

Example with increased `RCVBUF_SIZE ` buffer and no package loss:
image

Samahu:
## Related Issues & PRs
- related #10 
- related #240
- resolves #247 

## Summary of Changes
- Bump up ouster_client to 20231031 release
- Remove ouster-ros melodic from build checks since it is no longer compatible (due to spdlog)

## Validation
Ensure all functionality sustained with no drawbacks per testing procedures for ouster-ros

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -15,7 +15,6 @@ jobs:
       fail-fast: false
       matrix:
         ros_distro:
-          - melodic
           - noetic
     steps:
       - uses: actions/checkout@v3
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,13 +7,14 @@ ouster_ros v0.10.0
 
 ouster_ros(1)
 -------------
-* updated ouster_client to the release of ``20231031`` [v0.10.0]; changes listed below.
-* breaking: with this release the ouster-ros driver is no longer compatible with ROS melodic
-* breaking: publish PCL point clouds destaggered.
+* [BREAKING] updated ouster_client to the release of ``20231031`` [v0.10.0]; changes listed below.
+* [BREAKING]: with this release the ouster-ros driver is no longer compatible with ROS melodic
+* [BREAKING]: publish PCL point clouds destaggered.
 * introduced a new launch file parameter ``ptp_utc_tai_offset`` which represent offset in seconds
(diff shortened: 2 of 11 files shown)


Samahu:
So apparently the build fails on ouster_ros melodic because spdlog under Ubuntu 18.04 does not support spdlog formatting. This means we'd have to modify the SDK to not use this method or just disable ROS melodic altogether.

Samahu:
> So apparently the build fails on ouster_ros melodic because spdlog under Ubuntu 18.04 does not support spdlog formatting. This means we'd have to modify the SDK to not use this method or just disable ROS melodic altogether.

Given that the SDK drops support for the Ubuntu 18.04 on October (https://github.com/ouster-lidar/ouster_example/discussions/532) I don't see a good alternative but to also drop support for melodic (completely) starting with version 0.10.

Samahu:
Resolve all conflict and did performed sanity checks

bexcite (approved):

Issue #231, opened 2023-10-13
Consider changing `ptp_utc_tai_offset` default value


lowellausen:
**Describe your question**
A clear and concise description of your question.

Hello! Opening this issue here as an idea/suggestion :pray: 

Should  the default value for the new `ptp_utc_tai_offset` parameter (introduced here https://github.com/ouster-lidar/ouster-ros/pull/195)  be set to `0.0`?

We had our PTP synchronization working nicely (I think we followed this at some point https://github.com/ouster-lidar/ouster_example/issues/144), and after pulling the latest updates we had to look into the changelog to figure that we needed to change the aforementioned parameter from `-37.0` to `0.0` for our synchronization to work properly again.

My idea would be to set this to `0.0` so
1. It doesn't automatically "break" the system for people who pull the latest changes 
2. It removes the need to change all the `-37.0`s when a new leap second is added   

I can open a PR if we agree to some possible change on this issue!

Thanks for the attention!

**Platform (please complete the following information):**
- Ouster Sensor? OS-0 32 Rev D 
- Ouster Firmware Version?  2.5.2.
- ROS version/distro? Humble
- Operating System? Ubuntu 22.04 based ros:humble Docker image
- Machine Architecture? x64
- git commit hash: https://github.com/ouster-lidar/ouster-ros/pull/195

lowellausen:
Hi @Samahu, thanks for your answer! 

I understand your point and I agree, it is easy to adapt to this change as it is. With the upcoming breaking change indication in the changelog this will be quite clear! :ok_hand: 

Will close this one, since the proposed solution is underway. Thanks!

Hommus:
@Samahu I came across this issue and would like to suggest adding a note to the [Ouster PTP Quickstart Guide](https://static.ouster.dev/sensor-docs/image_route1/image_route2/appendix/ptp-quickstart.html) to help those in a similar situation in the future.

Samahu:
That's a good note. I will check with the page editor if they can add such a note.

PR #257, opened 2023-11-15
launch: Set default ptp offset to 0.0

ovaag:
## Related Issues & PRs
-
## Summary of Changes
`ptp_utc_tai_offset` is set to `-37.0` in all launch files on current master to reflect the current difference in seconds between TAI and UTC.

On my robot with ptp config already set up, this resulted in all my clouds to have a 37 sec offset from the rest of my system in `TIME_FROM_PTP_1588` mode.

## Validation
Ran driver with proposed changes and saw that clouds where in sync with rest of system.

### DIFF
--- a/launch/driver.launch
+++ b/launch/driver.launch
@@ -25,7 +25,7 @@
     TIME_FROM_PTP_1588,
     TIME_FROM_ROS_TIME
     }"/>
-  <arg name="ptp_utc_tai_offset" default="-37.0"
+  <arg name="ptp_utc_tai_offset" default="0.0"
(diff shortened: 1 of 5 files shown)


ovaag:
On a second thought this should be -37.0 on upstream master if users have not set up a `ptp4l.conf` file on their system.

Samahu:
@ovaag thanks for submitting the PR, but I as I noted in my response on #231, we are going to leave the default offset at `-37` and note in the changelog that it is breaking change. When updating to the newer version users should override the parameter through their launch command if they'd rather have a zero ptp/utc offset. The default offset works readily for new starters who have no clue about PTP/UTC time standards.

PR #258, opened 2023-11-15
Cmakelists: Fix install directories

ovaag:
This fixes issues of including headers from ouster_client or nonstd when using install space.

Folder layout  on current master:
image

Note e.g. `os_point.h` being located in `install/include/ouster_ros/include/ouster_ros/os_point.h`
instead of just `install/include/ouster_ros/os_point.h` where I would expect it to be.

How `install/include` looks after this PR:
image
image
and so on..

This means that if you are using e.g. `stagger` function from `lidar_scan.h` in you own ros_pkg, you can now:

```
#include <ouster/lidar_scan.h>
ouster::stagger(my_img,pixel_shift_vec)
```

## Related Issues & PRs
-
## Summary of Changes
Update install path in CmakeLists.txt
## Validation
Included header from ouster_client and compiled with install space. It works!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -123,9 +123,8 @@ install(
 
 install(
   DIRECTORY ouster-sdk/ouster_client/include/optional-lite/nonstd/
-  DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}/../nonstd
+  DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}/../ouster/nonstd/
 )
-
 install(
   FILES
     LICENSE
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -112,8 +112,18 @@ install(
 )
 
 install(
-  DIRECTORY ${_ouster_ros_INCLUDE_DIRS}
+  DIRECTORY include/ouster_ros/
   DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}
+)
+
+install(
+  DIRECTORY ouster-sdk/ouster_client/include/ouster/
+  DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}/../ouster
+)
+
+install(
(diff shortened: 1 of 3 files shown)


Samahu:
This makes sense but only concern is there were different dependencies (ROS packages) using the nonstd::optional library and they by default did the same thing, wouldn't they collide when installing to the same folder.

CMakeLists.txt
```diff
@@ -104,10 +104,20 @@ install(
...
+install(
+  DIRECTORY ouster-sdk/ouster_client/include/ouster/
+  DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}/../ouster
+)
+
+install(
+  DIRECTORY ouster-sdk/ouster_client/include/optional-lite/nonstd/
+  DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}/../nonstd
+)
+
```
ovaag, line 120: > This makes sense but only concern is there were different dependencies (ROS packages) using the nonstd::optional library and they by default did the same thing, wouldn't they collide when installing to the same folder.

I see your point. Do you think it would be okay to just not install it? We are primarily interested in the Ouster headers here.
```suggestion
```

Samahu: How about we just place it in the same directory under ouster? this way it is in the same folder as `types.h` which makes use of this header file and at the same we don't run into collision, so this would be my proposal:
```suggestion
install(
  DIRECTORY ouster-sdk/ouster_client/include/optional-lite/nonstd/
  DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}/../ouster/nonstd/
)
```

Samahu (changes requested):
Beside changing the path for `nonstd/optional.hpp` we would need the following to have this merged:
- [ ] increase package minor version (0.11.0 to 0.12.0)
- [ ] add a comment to the change logs in the unreleased section about this change

ovaag:
Rebased on latest master, increased package.xml minor version and added comment in changelog now.

Samahu (approved):

Issue #265, opened 2023-11-21
Ros parameter to not publish cloud if dropped udp packages


ovaag:
**Is your feature request related to a problem? Please describe.**
On our robot, the ouster driver sometimes output incomplete pointclouds where less than 32 lidar packets have been used to create the cloud. This leads to our odometry pipeline working on weird looking clouds and producing misleading scanmatching results.
In these cases we'd rather have no data, than to have bad data.

**Describe the solution you'd like**
A ros parameter that can be altered to tell the driver if you'd like it to not publish the pointcloud if it's not made up of all the expected lidar packages. 

**Describe alternatives you've considered**
Having some node doing filtering directly after ouster driver to detect sudden drops in number of points in cloud.

I'm currently looking into doing this on our fork of ouster-ros, so I was wondering if this is something you would have interest in having upstream as well? If so, do you have strong opinions on how it should be done, or should I just submit the PR, and take it from there?

Samahu:
It makes total sense

PR #268, opened 2023-11-23
Param min packets in cloud

ovaag:
## Related Issues & PRs
https://github.com/ouster-lidar/ouster-ros/issues/265

## Summary of Changes
This adds a parameter the user can alter in order to discard clouds that consists of x amount of lidar packets.

## Validation
I have an example bag where we experienced bad connectivity to lidar, resulting in dropped lidar packets.

Validated this PR by doing the following with replay of named bag:

Set param to 0, see that behaviour is as before:

https://github.com/ouster-lidar/ouster-ros/assets/50892582/e839fc11-c927-4c12-be15-a6afefffb2c3

Set param to 500 (I'm in lidar_mode 512x64)

https://github.com/ouster-lidar/ouster-ros/assets/50892582/1edc9f19-1f2a-4533-861d-bc8a80ddc874

corresponding terminal output:
image

### DIFF
--- a/launch/common.launch
+++ b/launch/common.launch
@@ -41,6 +41,15 @@
     xyzir
     }"/>
 
+  <arg name="min_lidar_packets_per_cloud" doc="Each cloud is built up of an expected
+  amount of udp packets. With this set to 0, the driver will publish all clouds regardless
+  of how many packets it received. If set to a positive integer, the driver will discard the entire
+  cloud if it does not consist of at least this many udp packets.
+  Number of expected packets is dependent on lidar_mode and udp_profile_lidar.
+  To calculate the number of expected packets, use the following formula:
+
+  num_expected_packets = info.format.columns_per_frame * info.format.columns_per_packet"
+  />
 
   <group ns="$(arg ouster_ns)">
     <node pkg="nodelet" type="nodelet" name="os_cloud_node"
@@ -60,6 +69,7 @@
       <param name="~/scan_ring" value="$(arg scan_ring)"/>
       <param name="~/ptp_utc_tai_offset" type="double" value="$(arg ptp_utc_tai_offset)"/>
       <param name="~/point_type" value="$(arg point_type)"/>
+      <param name="~/min_lidar_packets_per_cloud" value="$(arg min_lidar_packets_per_cloud)"/>
     </node>
   </group>
 
--- a/launch/common.launch
+++ b/launch/common.launch
@@ -41,15 +41,12 @@
     xyzir
     }"/>
 
-  <arg name="min_lidar_packets_per_cloud" doc="Each cloud is built up of an expected
-  amount of udp packets. With this set to 0, the driver will publish all clouds regardless
-  of how many packets it received. If set to a positive integer, the driver will discard the entire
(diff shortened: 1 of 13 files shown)


ovaag:
Note: My original implementation here was [d41c11e](https://github.com/ouster-lidar/ouster-ros/pull/268/commits/d41c11e1fb6a67d2f9d2fbe6c1e506ed58fe3fab) and [0aab35a](https://github.com/ouster-lidar/ouster-ros/pull/268/commits/0aab35a08d14994f891022490e5d398490910899).

I changed my approach to checking the condition in the nodelet, which felt cleaner.

Think the latter approach also will scale nicely if a user want to publish lidar_scan metadata e.g. shot_limiting or thermal shutdown status as this can now just be added to the OutputType struct

lmk what you think @Samahu

Samahu:
I had a brief look at your PR :eyes:  but the way I see it, it would be much better to implement this filter at the LidarPacketHandler level rather than having to process this unwanted scan and pass it to all registered processors and then drop all the byproducts before publishing.

Also what is the use case to have this parameter as a number rather than a boolean? Do you have a certain threshold where you actually use to discard the frame or do you usually set this value to `1`?

ovaag:
Hi again,

We have made some changes to our odometry pipeline that will handle partial lidar scans better. 
This means that the scope of this (or maybe a new PR?) can be narrowed.
We can drop all the code and parameter handling that drops sending pointclouds if they contain less than x % of expected columns. 

However, I'm still very interested in publishing the number of missing columns in a scan for diagnostics purpose. We've seen that this number can indicate some bad connection or network config. We would like to pick up problems like this as early as possible for our customers.

Would this be something that could possibly go into the new `Telemetry` msg? 

Also, if this is something that will just clutter your upstream code, let me know :+1:  Then I'll implementing the diagnostics on my fork. 

Regarding this:
> but the way I see it, it would be much better to implement this filter at the LidarPacketHandler level rather than having to process this unwanted scan and pass it to all registered processors and then drop all the byproducts before publishing.

Do you mean that it would be better to have the check here?

https://github.com/ouster-lidar/ouster-ros/blob/c5ce565c715f4673374bbaf13e81159d37e33b2e/src/lidar_packet_handler.h#L181-L187

I guess it would then look something like ?

```
        std::unique_lock<std::mutex> lock(*mutexes[ring_buffer.read_head()]);

        auto& lidar_scan = *lidar_scans[ring_buffer.read_head()]
        for (auto h : lidar_scan_handlers) {
            h(lidar_scan, lidar_scan_estimated_ts,
              lidar_scan_estimated_msg_ts);
        }
        const auto num_valid_columns = (lidar_scan.measurement_id().array() != 0).count() + 1;

```

How would you propose getting the `num_valid_columns` information back to the ros nodelets in a clean way? 
Looking forward to hearing from you on this :)

Samahu:
> Would this be something that could possibly go into the new Telemetry msg?

The `Telemetry` message is processed per packet, it does not have an overview of the `LidarScan`. However, I think with additional work one could potentially develop some stats of received scans and send them via the telemetry message (or a different stats message).

Samahu:
Mostly fulfilled by #428

PR #302, opened 2024-02-21
Atomic refactor for lock-free ringbuffer [Foxy]

Imaniac230:
## Related Issues & PRs
* https://github.com/ouster-lidar/ouster-ros/issues/240
* possibly related to https://github.com/ouster-lidar/ouster-ros/issues/272
* possibly related to https://github.com/ouster-lidar/ouster-ros/issues/280

## Summary of Changes

A proposed refactor of the `ThreadSafeRingBuffer` using `std::atomic` variables for lock-free access.

Changes:
1. Blocking `read` and `write` methods continue to rely on condition variables for buffer `full` and `empty` estimation. The locks are only held for the duration of `wait` and once `wait` is finished, the operations are performed atomically.
2. The `write_overwrite` method is now non-blocking, meaning it will atomically overwrite the last item imemdiately. This will not push the `read_idx` forward if an overwrite occurs.
3. The `read_timeout` method continues to rely on the condition variable for buffer `empty` estimation. Lock is only held for the duration of `wait_for`. If the `wait_for` is successful, it will perform the read atomically. If the `wait_for` timeouts, it will simply not perform the callback (instead of returning a `nullptr`) and unlock the lock.
4. Implemented non-blocking `write_nonblock` and `read_nonblock` methods, which return immediately and only perform the `read` or `write` operation if the `empty` and `full` conditions are met. These conditions are checked atomically.
5. The read and write indexes are atomically incremented before the the actual operation takes place, instead of after. This way we can for example assume that the writer currently has "ownership" of the data pointed to by the current `write_idx` and might be performing a write. If the index was incremented afterwards, it's current/actual value would be basically meaningless.

I've taken some liberty with the following changes, which may not be desired but make more sense in my limited understanding:
1. We will not be returning a `nullptr` in any case. If the corresponding conditions for a nonblocking or timeouted operation are not met we simply don't perform the callback operation and don't modify the buffer active size. I find it safer to just ignore the callback instead of returning an invalid pointer, which then must be handled safely by the caller.
2. The overwriting method will not push the reading index forward if the buffer is being overwritten. This is no longer desirable if both, the reading and writing thread modify the index asynchronously. The writer can keep overwriting the buffer as many times and as fast as it can, and the reader will simply continue at it's own pace. The only thing we have to make sure of is that they both never attempt to access the same field at the same time, which is assured by atomic access to the indexes.
3. To make sure the reader and writer will not attempt to acces the same index at the same time, there are two rules:
[...]

PR #321, opened 2024-04-16
Implement lock free ring buffer with throttling [HUMBLE/IRON]

### LINKED ISSUE #240: Missing lidar chunk os1-128-u
**Describe the bug**
We are using a Nvidia Jetson AGX xavier on a Auvidea X221-AI carriert board and we experience some missing pointcloud chunks on many of the pointclouds we receive. The missing chunk is always in the same spot, but sometimes larger and sometimes smaller. This happens on all udp profiles except for the Low Data Rate Profile. On the dual return profile it is more extreme. 

We are on commit 6a7693c551a34fc9f8dc47c8dda09cc64e2c57c0 and had nothing running in the background on a fresh ubuntu 20.04 install. The lidar is connected to a "RTL8111/8168/8411 PCI Express Gigabit Ethernet Controller". Lidar firmware version is v2.5.2

Here is a video of the behavior.

https://github.com/ouster-lidar/ouster-ros/assets/17435797/1f5c6a14-3ef4-468e-8e70-c9aff17282dd

**To Reproduce**
Steps to reproduce the behavior (steps below are just an example):
1. source ros environment
2. ouster-ros
3. source the project workspace
4. roslaunch ouster_ros driver.launch ...
5. observe the issue ...

**Platform:**
- Ouster Sensor: OS1-128-U
- Ouster Firmware Version: v2.5.2
- ROS version: noetic
- Operating System: Ubuntu 20.04
- Machine Architecture: arm64
- git commit: 6a7693c551a34fc9f8dc47c8dda09cc64e2c57c0

### LINKED ISSUE #272: Parts of point cloud missing
**Describe the bug**
Hi, We are using OS-0 32-channel Rev6 lidar, We have found instances, where significant chunks of the point cloud are missing. I have attached a screen recording below which shows the anomalous behaviour. 

I have already gone through the issue mentioned [here](https://github.com/ouster-lidar/ouster-ros/issues/240)
but I am not sure If the cause is the same.

There are 3 possible points of failures 
1) Power supply - We are querying the sensor's telemetry to check for the V, A, and temperature

- status: "\"RUNNING\""
- temperature: 35
- input_voltage_mv: 1151.0
- input_current_ma: 1759.0 

The usual values of current and voltage are mentioned above, can you please mention the exact cutoff values below which the sensor would stop functioning, the docs mention 12V but even with Ouster's provided power supply we didn't get telemetry reporting 12V

2) Ethernet Switch - Can you point out some methods to detect packet loss or bandwidth throttling to isolate if the ethernet switch is the culprit?

3) Driver - Any method to isolate if there are any issues related to the driver?

**Setup**
The Lidar and the edge device are connected to a Gigabit Ethernet switch and the lidar is connected to a constant 12V power supply

- Ouster Sensor \[OS-0\]
- Ouster Firmware Version \[Latest\]
- Programming Language \[C++\]
 
- Operating System \[Ubuntu 18.04\]
- Machine Architecture \[arm]

https://github.com/ouster-lidar/ouster-ros/assets/51359628/cf276b37-f113-4f57-92a7-48a5c0f63cee

Samahu:
## Related Issues & PRs
- Closes #240
- Closes #272
- Related #302
- ROS1[NOETIC] #319 
- ROS2[FOXY] #320 

## Problem Analysis
- When adding multiple subscribers or when having a bad network or low performing system, LidarScan composition and processing could eventually take more time, this leads to a build up in the packets buffer which will eventually be exhausted and the current mechanism of the thread_safe_ring_buffer::write_overwrite made things only worse. The solution can be in one of two things either ensure that the LidarScan composition + processing takes less time than the time needed to add new packets, or decouple packet addition from LidarScan consumption, if the consumption thread can't keep up then throttle. 

## Summary of Changes
- Implement lock free ring buffer with unit tests
- Decouple packet reception from packet processing
- Throttle the ring buffer as we reach 70% capacity
- Drop packets only when the ring buffer has been completely exhausted

#### TODO(s):
- Replace the **LockFreeRingBuffer** implementation with **ThreadSafeRingBuffer** within `os_sensor_nodelet`.
  - Alternatively, forward the packets directly to LidarScan without a staging buffers.  

## Validation
- Run the driver as usual with RIVZ, add multiple subscribers to `/ouster/points`
- Verify the problem with missing columns is eliminated or reduced.
- Observe that the point cloud publish frequency isn't affected significantly

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,6 +8,7 @@ Changelog
   * address an issue where LaserScan appeared different on FW prior to 2.4
 * [BUGFIX]: LaserScan does not work when using dual mode
 * [BUGFIX]: ROS2 crashes when standby mode is set and then set to normal
+* [BUGFIX]: Implement lock free ring buffer with throttling to reduce partial frames
 
 
 ouster_ros v0.12.0
--- a/ouster-ros/CMakeLists.txt
+++ b/ouster-ros/CMakeLists.txt
@@ -169,8 +169,10 @@ rclcpp_components_register_node(os_driver_component
 if(BUILD_TESTING)
   find_package(ament_cmake_gtest REQUIRED)
   ament_add_gtest(${PROJECT_NAME}_test
-    test/ring_buffer_test.cpp
     src/os_ros.cpp
+    test/test_main.cpp
+    test/ring_buffer_test.cpp
+    test/lock_free_ring_buffer_test.cpp
     test/point_accessor_test.cpp
     test/point_transform_test.cpp
     test/point_cloud_compose_test.cpp
--- a/ouster-ros/package.xml
+++ b/ouster-ros/package.xml
@@ -2,7 +2,7 @@
 <?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
 <package format="3">
   <name>ouster_ros</name>
-  <version>0.12.2</version>
+  <version>0.12.3</version>
   <description>Ouster ROS2 driver</description>
   <maintainer email="oss@ouster.io">ouster developers</maintainer>
   <license file="LICENSE">BSD</license>
--- a/ouster-ros/src/lidar_packet_handler.h
+++ b/ouster-ros/src/lidar_packet_handler.h
@@ -16,6 +16,9 @@
 
 #include <pcl_conversions/pcl_conversions.h>
 
+#include "lock_free_ring_buffer.h"
+#include <thread>
+
 namespace {
 
 template <typename T, typename UnaryPredicate>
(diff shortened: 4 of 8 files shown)


TillBeemelmanns:
We tested both, this branch and this attempt https://github.com/ouster-lidar/ouster-ros/pull/302 to solve the problem with missing parts of the point cloud. We set the recv buffer size to the following values 
```bash
sudo sysctl -w net.core.rmem_default=26214400
sudo sysctl -w net.core.rmem_max=26214400
```
on the host machine and also set `proc_mask:="PCL"`. However in all cases we could observe partial missing point clouds as soon as we add a second subscriber node to the driver. That applied to resolutions `1024x20, 2048x10, 4096x5`, for the other resolutions `512x10,  512x20, 1024x10` we could observed a significant delay in the point cloud. 

For all resolutions, the point cloud looks fine if only RVIZ is used for visualization (single subscriber).

marioney:
We also tested this PR with an `OS-1-128 Rev 7` and it fixed the missing points issue for us, running ROS Humble in Ubuntu 22.04.

We tested both 1024x10 and 1024x20 modes and the issue was not present in any of those.

Samahu:
@TillBeemelmanns  @marioney thank you both for the feedback.

@TillBeemelmanns sorry that didn't help with your situation .. I still have one TODO item that I predict it to help with the delay and might improve things overall. I should have this PR completed and merged by next week.

mkiael:
Tested on `OS0-128 Rev C` running ROS Humble on Ubuntu 22.04. Solves issues with missing data.

Great work!

Samahu:
Thanks @mkiael for the feedback. I will proceed with the PR as is since it resolves the issue for the most and will address the rest of improvements separately.

James-R-Han:
Hi @Samahu!

Following up on this thread, similar to @TillBeemelmanns, after the PR, I still encounter a drop of about ~40% of the point cloud when I have 2 subscribers and no drop when it's only the rviz (single subscriber). Increasing the buffer unfortunately did not solve this problem. I am also running with minimized bandwidth configurations:  point_type=xyzi, proc_mask="PCL", and UDP Profile Lidar = RNG15_RFL8_NIR8

Are there any updates regarding this problem?

James-R-Han:
The issue does not occur with FastRTPS; CycloneDDS has this dropping issue. But as per other discussions, using FastRTPS, we can not hit 10Hz for the point cloud topic. 

Also I have tried the galactic branch with CycloneDDS and I encounter the same issue.

Samahu:
I do have further improvements that I want to implement which should help with the situation but can't give you an ETA yet.

PR #330, opened 2024-05-25
Add parameter to optionally disable the publication  of the static tf

doisyg:
## Summary of Changes

This PR adds the parameter `pub_static_tf`, by default `True`, to control if the user wants or not to have the driver node publishing the static transform between the sensor frames.
It is useful to have it disabled in systems where a calibration process directly modifies the tf of the point cloud frame.


nachovizzo:
@Samahu it would be great if we could discuss moving forward with this PR as it allows to have a little bit more control over how to manipulate the static transformations of the 3 different ouster coordinate frames

Samahu:
@nachovizzo Sure, I will prioritize the merge of @doisyg work within couple days

Samahu (approved):
merging into a branch so I can add expose the new parameter to config

PR #356, opened 2024-08-19
ROS-227: Set LIDAR FOV on startup and add an option to persist the config [NOETIC]

### LINKED ISSUE #138: Support setting sensor config
I have to set sensor config separately, like for example Azimuth window.

- on configure, read the config
- if the config doesn't match the params passed, write new config

Writing a separate program or script to configure the azimuth window

- Ouster Sensor? N/A
- Ouster Firmware Version? N/A
- ROS version/distro? ROS 2
- Operating System? N/A
- Machine Architecture? N/A

### LINKED ISSUE #227: Changing the FOV of the Lidar in ROS
I want to change the FOV of the Ouster OS0-32 from [0-360] to [0, 180000] and in the Software User guide i see  the use of ``` set_config_param azimuth_window``` How can i set this param in ros? I don't see this particular rosparam in rosparam list.

**Platform :**
- Ouster Sensor: OS-0-32
- ROS version/distro Melodic
- Operating System? Ubuntu 18

Samahu:
## Related Issues & PRs
- closes #138
- related #227

## Summary of Changes
* Set LIDAR FOV on startup
* Add an option to persist the config

## Validation
* Run the node with azimuth start and stop as follows 
```bash
roslaunch ouster_ros driver.launch \
    azimuth_window_start:=150000 \
    azimuth_window_end:=250000
```
Verify that the azimuth window is cropped to selection 

* Run the node with `persist_config` set **true** as follows 
```bash
roslaunch ouster_ros driver.launch \
   <pass a selection of settings> \
    persist_config:=true
```
Verify that the new settings persist after power up/down cycle of the sensor

### DIFF
--- a/launch/driver.launch
+++ b/launch/driver.launch
@@ -66,9 +66,9 @@
     }"/>
 
   <arg name="azimuth_window_start" default="0" doc="azimuth window start;
-    values range [0, 360000] millideggrees"/>
+    values range [0, 360000] millidegrees"/>
   <arg name="azimuth_window_end" default="360000" doc="azimuth window end;
-    values range [0, 360000] millideggrees"/>
+    values range [0, 360000] millidegrees"/>
 
   <arg name="persist_config" default="false"
     doc="request the sensor to persist settings"/>
--- a/launch/driver.launch
+++ b/launch/driver.launch
@@ -70,6 +70,9 @@
   <arg name="azimuth_window_end" default="360000" doc="azimuth window end;
     values range [0, 360000] millideggrees"/>
 
+  <arg name="persist_config" default="false"
+    doc="request the sensor to persist settings"/>
+
   <group ns="$(arg ouster_ns)">
     <node pkg="nodelet" type="nodelet" name="os_nodelet_mgr"
       output="screen" required="true" args="manager"/>
@@ -99,6 +102,7 @@
       <param name="~/point_type" value="$(arg point_type)"/>
       <param name="~/azimuth_window_start" value="$(arg azimuth_window_start)"/>
       <param name="~/azimuth_window_end" value="$(arg azimuth_window_end)"/>
+      <param name="~/persist_config" value="$(arg persist_config)"/>
     </node>
   </group>
 
--- a/launch/driver.launch
+++ b/launch/driver.launch
@@ -65,6 +65,11 @@
     xyzir
     }"/>
 
+  <arg name="azimuth_window_start" default="0" doc="azimuth window start;
+    values range [0, 360000] millideggrees"/>
+  <arg name="azimuth_window_end" default="360000" doc="azimuth window end;
+    values range [0, 360000] millideggrees"/>
+
   <group ns="$(arg ouster_ns)">
     <node pkg="nodelet" type="nodelet" name="os_nodelet_mgr"
       output="screen" required="true" args="manager"/>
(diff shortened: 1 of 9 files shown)


src/os_sensor_nodelet.cpp
```diff
@@ -313,6 +317,13 @@ sensor::sensor_config OusterSensor::parse_config_from_ros_parameters() {
         config.udp_port_imu = imu_port;
     }
 
+    persist_config = nh.param("persist_config", false);
+    if (persist_config && (lidar_port == 0 || imu_port == 0)) {
+        NODELET_WARN("When using persist_config it is recommended to not "
+        "use 0 for port value");
+    }
```
Samahu, line 324: I am not sure if this warning would be necessary, open to your suggestions

launch/driver.launch
```diff
@@ -65,6 +65,14 @@
     xyzir
     }"/>
 
+  <arg name="azimuth_window_start" default="0" doc="azimuth window start;
+    values range [0, 360000] millideggrees"/>
+  <arg name="azimuth_window_end" default="360000" doc="azimuth window end;
+    values range [0, 360000] millideggrees"/>
```
Samahu, line 71: Using a default of `0` and `360000` can be breaking change for some users since it will reset their current Azimuth Window setting. We can avoid this by using `-1` or any negative value as to no change, or use a empty string value similar what other options we had.

Krishtof-Korda (approved):

launch/driver.launch
twslankard: Nit, spelling: `millidegrees`

src/os_sensor_nodelet.cpp
matthew-lidar: If only the FW would let you change ports without a reinit, then all our problems would go away.....

I think the warning is fine, maybe could make the language more specific.

"When using persist_config it is recommended to not use 0 for port value as this requires reinitializing the sensor each time anyways."

Samahu: > If only the FW would let you change ports without a reinit, then all our problems would go away.

Unfortunately it doesn't allow that :face_with_diagonal_mouth:

matthew-lidar: Maybe we can use this as an example for making the case to allow that

matthew-lidar (approved):
Looks good, just maybe consider changing the language of the warning.

src/os_sensor_nodelet.cpp
twslankard: > As a crazy idea, it is probably possible to look at the current sensor port, see if it is available and just use that rather than grabbing a new ephemeral port. That way we can mostly avoid reinits.

I like this suggestion.

twslankard (approved):
LGTM ! :rocket:

src/os_sensor_nodelet.cpp
Samahu: > As a crazy idea, it is probably possible to look at the current sensor port, see if it is available and just use that rather than grabbing a new ephemeral port. That way we can mostly avoid reinits.

It isn't crazy to me, however, this needs further work. I'd suggest to achieve this in a follow up ticket and have the core sdk provide this functionality (check if the current ports accessible) then we can use here.

PR #359, opened 2024-08-23
Link pcap nodelet only if BUILD_PCAP is defined

ovaag:
## Related Issues & PRs
When using current master with `BUILD_PCAP` set to default value `OFF` i got the following error:
image

## Summary of Changes
This seems to stem from that pcap_node is linked into the nodelets library regardless of the `BUILD_PCAP` cmake arg

## Validation
Tested on my robot with ouster lidar. `roslaunch ouster_ros sensor.launch <my_args>` now runs as expected


Samahu:
Hi @ovaag thanks for reporting the issue, somehow I can't reproduce the problem on my machine even though I made sure to clean the workspace and rebuilt with pcap off. But it does make sense to me to not include `os_pcap_nodelet.cpp` in the `ouster_ros_nodelets` lib

Samahu (approved):

Issue #9, opened 2021-08-13
Ouster Time Handling for tf and sensor msgs 


tahir1069:
I have a question about Ouster driver handling for ROS.  I was having an error in `tf lookupTransform` that the transform is old. Ouster driver publishing the `tf` is [using ros time](https://github.com/ouster-lidar/ouster_example/blob/master/ouster_ros/src/ros.cpp#L107) where as other sensor data like lidar and imu are using [here](https://github.com/ouster-lidar/ouster_example/blob/master/ouster_ros/src/ros.cpp#L95) and [here](https://github.com/ouster-lidar/ouster_example/blob/master/ouster_ros/src/ros.cpp#L37).

I know Ouster can handle different timestamp methods, but I don't understand the reason of using it for sensor readings but not for `tf`.

VictorJCS:
Hi @Samahu 
I have the same problem, I am using OS0-64-U and Rtabmap package (used for SLAM). Actually, in normal conditions the ouster node and Rtabmap node running without problem.

![image](https://user-images.githubusercontent.com/85325944/201322453-7e71fed5-5553-4e33-ab32-10b73c5f79e3.png)

But, I have problems when I try to do a test:
1. Ouster node and Rtabmap is running
2. I disconnect Lidar sensor
3. Ouster node and Rtabmap node continue running without data
4. After 2 minutes I connect Lidar sensor

The nodes reprocess Lidar data but I have the following error:

[ WARN] [1668158414.957718441]: odometry: Could not get transform from base_link to os_sensor (stamp=1668158375.848681) after 0.500000 seconds ("wait_for_transform_duration"=0.500000)! Error="Lookup would require extrapolation 38.615832763s into the past.  Requested time 1668158375.848681211 but the earliest data is at time 1668158414.464514017, when looking up transform from frame [os_sensor] to frame [base_link]. canTransform returned after 0.504406 timeout was 0.5."

The point cloud is like it:

![image](https://user-images.githubusercontent.com/85325944/201318103-ac82f7e4-825b-4a42-8089-23d9ea64d608.png)

I am using the new release https://github.com/ouster-lidar/ouster_example/pull/420 and https://github.com/ouster-lidar/ouster_example/pull/421. I use TIME_FROM_ROS_TIME as timestamp_mode.

[ INFO] [1668158321.969738893]: Loading nodelet /os_cloud_node/os_node of type nodelets_os/OusterSensor to manager os_nodelet_mgr with the following remappings:
[...]

PR #361, opened 2024-08-23
Implement sensor reconnection behavior

Samahu:
## Related Issues & PRs
- includes https://github.com/ouster-lidar/ouster-ros/pull/119

## Summary of Changes
- Implement sensor reconnection behavior

## Validation
TBD

### DIFF
--- a/src/os_sensor_nodelet.cpp
+++ b/src/os_sensor_nodelet.cpp
@@ -27,6 +27,7 @@ using namespace std::chrono_literals;
 using namespace std::string_literals;
 using std::to_string;
 
+
 namespace ouster_ros {
 
 OusterSensor::~OusterSensor() {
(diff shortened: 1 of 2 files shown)


src/os_driver_nodelet.cpp
```diff
@@ -38,18 +38,66 @@ class OusterDriver : public OusterSensor {
...
+        // NOTE: always create the 2nd topic
+        lidar_pubs.resize(2);
+        for (int i = 0; i < 2; ++i) {
+            lidar_pubs[i] = getNodeHandle().advertise<sensor_msgs::PointCloud2>(
+                topic_for_return("points", i), 10);
+        }
+    }
+
+    void create_laser_scan_pubs() {
+        // NOTE: always create the 2nd topic
```
Samahu, line 75: We used to create the 2nd topics only if the current profile supported DUAL returns, however, since the user can potentially switch between different profile on reconnection it wouldn't bound well that we create and destroy topics when switching between dual vs non-dual for downstream consumers.

src/os_cloud_nodelet.cpp
```diff
@@ -72,53 +73,74 @@ class OusterCloud : public nodelet::Nodelet {
...
         auto proc_mask = pnh.param("proc_mask", std::string{"IMU|PCL|SCAN"});
         auto tokens = impl::parse_tokens(proc_mask, '|');
 
         auto timestamp_mode = pnh.param("timestamp_mode", std::string{});
         double ptp_utc_tai_offset = pnh.param("ptp_utc_tai_offset", -37.0);
 
-        auto& nh = getNodeHandle();
-
         if (impl::check_token(tokens, "IMU")) {
-            imu_pub = nh.advertise<sensor_msgs::Imu>("imu", 100);
```
matthew-lidar, line 99: It seems like you are always creating the publishers now. Do you mean to do that?

If so we should get rid of the proc_mask and just detect if one of the topics are subscribed to before generating/publishing it.

src/os_sensor_nodelet.cpp
```diff
@@ -39,26 +40,78 @@ void OusterSensor::halt() {
...
+    // deactivate
+    halt();
+    sensor_client.reset();
+}
+
+void OusterSensor::attempt_start() {
+    if (!start()) {
+        if (attempt_reconnect) {
+            reconnect_timer = getNodeHandle().createTimer(
+                ros::Duration(3.0), [this](const ros::TimerEvent&) {
```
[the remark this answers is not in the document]
matthew-lidar: Probably. It is a magic number. I wouldn't go through the trouble of putting it in launch files though.

src/os_cloud_nodelet.cpp
Samahu: I actually meant to create the topics only if their respective token was set. But forgot to add a check before creating the respective topic for each processor.

Samahu: addressed

src/os_sensor_nodelet.cpp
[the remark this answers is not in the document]
Samahu: yeah it isn't too difficult to add later on if some user really wanted to have some control over it. Let's skip it for now.

src/os_cloud_nodelet.cpp
```diff
@@ -72,53 +81,74 @@ class OusterCloud : public nodelet::Nodelet {
             tf_bcast.broadcast_transforms(info);
         } else {
             NODELET_INFO_STREAM(
-                "OusterCloud: dynamic transforms broadcast enabled wit "
+                "OusterCloud: dynamic transforms broadcast enabled with "
                 "broadcast rate of: "
                 << dynamic_transforms_rate << " Hz");
+            timer_.stop();
```
matthew-lidar, line 87: Is this actually necessary? Mostly curious

Samahu:
I think there was some bug in melodic where this was causing an issue on
for users that simply using the static broadcast is giving errors about out
of sync. I wish to remove it but Is it still needed .. not sure :shrugging:

src/os_cloud_nodelet.cpp
Samahu: There was a bug of some sort
https://github.com/ouster-lidar/ouster-ros/issues/9#issuecomment-1311528903 that was causing an issue on for users (including me) that simply using the static broadcast is giving errors about out of sync. I wish to remove it personally :shrugging:

twslankard (approved):

src/os_cloud_nodelet.cpp
Samahu: It seems I maybe have misunderstood you question .. it probably isn't necessary since I am recreating the timer.

celentes:
Would be good to have max_reattempts parameter for some cases - it makes sense to have when sensor connection has been dropped, may not make sense to have when first spinning up the sensor though.

celentes (approved):

src/os_sensor_nodelet.cpp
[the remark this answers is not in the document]
Samahu: added a launch arg to configure

Issue #352, opened 2024-08-06
"DROPPING PACKET" and "70% full, THROTTLING"


James-R-Han:
Hello!

I am using

- OS0-128 
- ROS2 Humble (with CycloneDDS) 
- UDP Profile RNG15_RFL8_NIR8
- Run configuration : `ros2 launch ouster_ros sensor.launch.xml sensor_hostname:=192.168.131.18 use_system_default_qos:=true timestamp_mode:=TIME_FROM_ROS_TIME sensor_qos_profile:=reliable proc_mask:="IMG|PCL" viz:=false`

I will experience the following errors:
![image](https://github.com/user-attachments/assets/023b2a31-a85d-423b-a394-b2ed2faed341)

Any advice on how to handle the situation?

Thank you in advance!

kavishmshah:
Hi, 
I too see the same when I have multiple subscribers to the pointcloud topic.
I'm using an nvidia Orin devkit and OS0-128U with 512x10, I don't see this issue occur but with 1024x10, I see packet dropa when Rviz is open and I record a bag simultaneously. 

What resolution are you using?

James-R-Han:
Hey Kavish! I'm glad I'm not the only one haha. I was using 1024x10 when I got the messages above

Samahu:
Hi @James-R-Han and @kavishmshah thanks for sharing the feedback.

I noticed that `sensor_qos_profile:=reliable` which suggest that you are not using the BEST_EFFORT which is what used for `SensorDataQoS`. I think choosing the RELIABLE QoS for the sensor will increase the holdup on the published queue resulting in the throttling and dropped packets. I recommend that you use the default `SensorDataQoS` for live processing and only use RELIABLE when capturing data.

In any case, as I noted in the current merged fix https://github.com/ouster-lidar/ouster-ros/pull/321 I have further TODOs that I want to implement soon which should improve the performance on the driver side.

So stay tuned :crossed_fingers:

James-R-Han:
Thanks @Samahu! 

When I've tried BEST_EFFORT publisher with BEST_EFFORT subscriber (RVIZ), empirically I see the frame rate will occasionally drop well below 10Hz.

When I use RELIABLE publisher with BEST_EFFORT subscriber, sometimes the RVIZ is just blank (see below). 

If I use RELIABLE for both I find the best result - the frame rate is consistently high; I'm able to wave my hand around and it's a smooth motion in the point cloud.

[...]

Issue #306, opened 2024-03-02
Lidar points dropped in ros bag record


yashmewada9618:
Hello,

I am trying to record a ros bag for Ouster lidar and an Intel Real sense camera including their IMU data and I am splitting the bag files every 3.2GB size margin. But after I do `ros2 bag info <bag_file>` I see that I have lost almost 800 messages for 92s of the bag file which is not good, even the IMU data is lost. 

Everything is working perfectly offline without rosbag record and I am receiving data at the given Hz and perfect bandwidth.  After following the changes from #302 the results were somewhat better but not perfect. I have also tried the steps from all the related issues here and they somewhat solved the problems I have been facing for over so long but didn't solve completely.

The lidar seems to drop messages as the number of subscribers increases. Has anyone faced this kind of issue or am I missing something here?

I've included the details below.

Thanks.

**To Reproduce**
**Ouster Lidar Launch**
```python
    ouster_launch = IncludeLaunchDescription(
        XMLLaunchDescriptionSource(
            [
                PathJoinSubstitution(
                    [FindPackageShare("ouster_ros"), "launch", "sensor.launch.xml"]
                )
            ]
        ),
        launch_arguments={
            "viz": "false",
            "sensor_hostname": "os-122212000733.local",  # Replace with the actual sensor hostname
            "lidar_mode": "1024x20",
            "udp_profile_lidar": "RNG19_RFL8_SIG16_NIR16",
            "metadata": "/home/scout/autonomy-stack/src/os-122212000733-metadata.json",  # Replace with relative path to metadata file
            "timestamp_mode": "TIME_FROM_ROS_TIME",
        }.items(),
    )
```
```python
# No compression of bag file, just splitting.
    rosbag_record = ExecuteProcess(
        cmd=[
            "ros2",
            "bag",
            "record",
            "-o",
            directory,
            "--qos-profile-overrides-path",
            Qos_override,
            "--max-cache-size",
            "1000000000",  # 1 GB
            "--max-bag-size",
            "3221225472",  # 3Gb of raw data
            "--max-bag-duration",
            "1800",  # seconds
            "-s",
            "sqlite3",
            "-a",
        ],
        shell=True,
    )
```

**Problem**

```bash
[ros2-10] 	/camera/camera/color/metadata: 329
[ros2-10] 	/ouster/range_image: 219
[ros2-10] 	/camera/camera/gyro/metadata: 2184
[ros2-10] 	/ouster/nearir_image: 219
[...]

Issue #259, opened 2023-11-16
Image gaps while LidarScan from LidarPackets


rmbsmit:
**Describe your question**
I have recorded a bagfile containing LidarPackets (type PacketMsg.msg). From this data, i am trying to batch them, recreate a LidanScan using the ScanBatcher and view the Intensity/Reflectivy/Range images in python. However, when looking at the final images, there seems to be a problem in either the batching of the LidarPackets, or in the destaggering. The Range output image looks like this: 

![25](https://github.com/ouster-lidar/ouster-ros/assets/48470663/b670e07b-790e-438d-8011-a07bcb53654d)

The image itself seems fine, however there are gaps in between the packages. 

The sample code look like this: 
```
        self.batcher = client._client.ScanBatcher(self.metadata)
        self.ls = client.LidarScan(height,
                          width,
                          self.metadata.format.udp_profile_lidar)

def Callback(self, msg):
packet = client.LidarPacket(msg.buf, self.metadata, _raise_on_id_check=True)
if self.batcher(packet, self.ls):
            ref_field = self.ls.field(client.ChanField.SIGNAL)
            ref_val = client.destagger(self.metadata, ref_field, inverse=False)
            ref_img = (ref_val / np.max(ref_val) * 255).astype(np.uint8)
```

**Platform (please complete the following information):**
- Ouster OS2
- ROS version: noetic
- Operating System: Ubuntu 20.04

Can somebody help me solve the issue?

Samahu:
Do you get the same gaps when replaying the bagfile through the driver?

rmbsmit:
No, replaying the bagfile, the images look fine.

![Screenshot from 2023-11-17 09-56-16](https://github.com/ouster-lidar/ouster-ros/assets/48470663/0a1065d9-0137-4b86-aa02-3271f33dab7d)

rmbsmit:
@Samahu do you have any update on this issue?  Thanks in advance

Samahu:
I don't have an update but looking briefly at your code I am wondering if you actually filter out ImuPacket before trying to parse the packet as a LidarPacket.

bexcite:
Hi @rmbsmit,

There are two things that I would check:

1) BAG lidar_packets content has no "holes" - i.e. all lidar_packets were received without drops and the network/system wasn't experiences the load or other form of disruptions that prevented the HOST machine to reveive all packets.
(to verify it you can use `ptudes viz` tool that I'll descrive below)

[...]

Issue #175, opened 2023-07-27
Reduce Vertical resolution for improving performance of the driver


shubham-shahh:
**Describe your question**
Hi, I am using OS0-128 lidar (REV-D) with the latest version of the driver. When the resolution is set to 512x10 and the packets are set to Legacy, I am receiving data at ~6-7Hz. but, when I am using an OS0-32 lidar (Rev-6), I am getting the data at the desired rate. I am willing to make sure I get data at 10Hz at 512x10 config. how can I reduce the amount of data processed by the driver? either by limiting the number of rings to 32 or reducing the points in some other practical ways.

**Platform:**
- Ouster Sensor? OS-32
- Ouster Firmware Version? V.2.5.2
- ROS version/distro? Melodic
- Operating System? Linux
- Machine Architecture? Arm

shubham-shahh:
> Have you switched to using `driver.launch` from `sensor.launch`?

I am using sensor.launch, but I saw the "scan_ring" param in driver.launch, I will try using driver.launch and update the issue

thanks

shubham-shahh:
Hi @Samahu, I looked through the code, I am looking to reduce the vertical resolution of the Pointcloud from let's say, 128 rings to 32 rings, I was unable to do that with the existing implementation.

thanks

shubham-shahh:
> @shubham-shahh please test #191 which shows how to reduce the vertical resolution.

Hi, thanks for quick solution, I'll test and update the thread.

thanks

shubham-shahh:
Hi @Samahu I tried the changes you mentioned, I am getting 32 rings as the output, but in the output cloud, the rings look distorted, below I have attached a screenshot for reference.
thanks
![distorted_rings](https://github.com/ouster-lidar/ouster-ros/assets/51359628/a4564da9-01d8-4a1e-a9f7-b6129921dc9f)

shubham-shahh:
> Hi @Samahu I tried the changes you mentioned, I am getting 32 rings as the output, but in the output cloud, the rings look distorted, below I have attached a screenshot for reference.
> thanks
> ![distorted_rings](https://github.com/ouster-lidar/ouster-ros/assets/51359628/a4564da9-01d8-4a1e-a9f7-b6129921dc9f)
> 

Hi @Samahu any update on this?

Samahu:
[...]

PR #363, opened 2024-09-04
[NOETIC] Further reduce overhead in the ros driver

### LINKED ISSUE #10: Loss lidar packet messages (ROS melodic)

### LINKED ISSUE #152: Filter out points below/above a certain range
Is there currently a convenient way to filter points below certain threshold, in a similar manner than the `min_range` parameter in `velodyne_driver`?

The sensor is often positioned so that the laser beams hit parts of the vehicle itself. This might disturb odometry/localization/SLAM algorithms, which think that these static points are part of the observed environment. This calls for having a functionality that could neglect all points that are e.g. at less than 3 meters distance from the sensor frame.

Implementing filters on the `PointCloud2` topic could be a hassle in terms of getting them work at a reasonable performance, so I think it would preferably need to be implemented in the driver, closer to the original data format, not on the xyz pointcloud ouput.

### LINKED ISSUE #238: `RAW` option for proc_mask
At the moment I am switching back and forth between the different XML and Python launch files. As far as I understand the preferred approach in the future would be to use `driver.launch.py`. The problem for me is that it does not publish raw lidar_packets which I like to record.

Would it be possible to add another key to proc_mask like `RAW` to publish raw lidar_packets?

At the moment I simply use `sensor.independent.launch.py`. (`sensor.independent.launch.xml` gave me recently some errors as described in this issue #159.)

- ROS version/distro? ROS2 humble
- Operating System? Jammy
- Machine Architecture? ARM

### LINKED ISSUE #251: Bring back optional publishing of lidar and imu packets to new single nodelet driver
We recently upgraded to use the new single nodelet driver (started via driver.launch as described in https://github.com/ouster-lidar/ouster-ros#sensor-mode ). This shows significantly reduced CPU consumption on Jetson embedded systems compared to the "old" multi-nodelet setup. However, it appears the support to simultaneously publish the lidar and imu packets has been dropped. This is a problem because for data recording purposes, those are the most useful data to record (as also remarked by others in https://github.com/ouster-lidar/ouster-ros/issues/110 ).

It would be great to have the single driver nodelet also optionally (i.e. `if numSubscribers() > 0` ) publish the raw packets, so they can be recorded for later postprocessing/analysis.

An alternative is using the old multi-nodelet setup, but that one causes high CPU load on Jetson embedded systems, so not a good option.

- ROS1, but likely applies to ROS2 also (?)

### LINKED ISSUE #304: deprecated declarations
**Describe the bug**
When compiling the ros2 branch with Ubuntu 22.04 & ros2 humble, gcc (11.4) prints several deprecated declaration warnings

**To Reproduce**
1. clone this repo and its submodules to a ros2 workspace
2. source /opt/ros/humble/setup.bash
3. colcon build --packages-up-to ouster_ros --event-handler console_direct+

**Screenshots**

- Ouster Sensor? not relevant
- Ouster Firmware Version? not relevant
- ROS version/distro? humble
- Operating System? Ubuntu Linux 22.04
- Machine Architecture? x86-64
- git commit hash (if not the latest). 11287a9

Samahu:
## Related Issues & PRs
- resolves #10 
- resolves #152 
- related #175 
- resolves #238 
- resolves #251 
- related #259
- resolves #304
- related #306 
- related #352

## Summary of Changes
- Add Jazzy to the build status
- Update to SDK client 0.11.1
- Switch to using LidarPacket and ImuPacket
  - Get's rid of the deprecation warnings
- Publish raw packets for the `os_driver`
- Add a flag to control the publish of organized point cloud vs non-organized
- Add a flag to control the publish of destaggerd point cloud vs staggered
- Add parameter to set the min/max lidar range on generated point clouds
- Add a new flag to `proc_mask` to optionally publish the raw packets when using os_driver

### TODO(s)
- ~optionally publish raw packets in os_driver~
- publicize the use of OpenMP in `cartesianT` & `point_cloud_compose` methods

## Validation
- Run the driver in all different modes of operations
- Verify that the newly added features:
  - When `destagger` is set to false verify the driver doesn't destagger the published point cloud
  - When `organized` is set to false verify that the published point cloud isn't 2D
  - Adjust `min_range` and `max_range` and observe the point cloud shrink accordingly.
- Verify that `/ouster/imu_packets`  and `/ouster/lidar_packet` appear when RAW flag is passed to `os_driver`


launch/sensor.launch
```diff
@@ -90,6 +93,16 @@
...
          Counter resets upon successful connection"/>
 
+  <arg name="organized" default="true"
+    doc="generate an organzied point cloud"/>
+  <arg name="destagger" default="true"
+    doc="enable or disable point cloud destaggering"/>
+
+  <arg name="min_range" default="0.0"
+    doc="minimum lidar range to consider (meters)"/>
+  <arg name="max_range" default="1000.0"
```
matthew-lidar, line 103: Why 1000? Can we set negative as unlimited or something? Seems weird to have an artificial limit

src/os_image_nodelet.cpp
```diff
@@ -52,7 +52,14 @@ class OusterImage : public nodelet::Nodelet {
     void create_lidar_packets_subscriber() {
         lidar_packet_sub = getNodeHandle().subscribe<PacketMsg>(
             "lidar_packets", 100, [this](const PacketMsg::ConstPtr msg) {
-                if (lidar_packet_handler) lidar_packet_handler(msg->buf.data());
+                if (lidar_packet_handler) {
+                    // TODO[UN]: this is not ideal since we can't reuse the msg buffer
+                    // Need to redefine the Packet object and allow use of array_views
+                    sensor::LidarPacket lidar_packet(msg->buf.size());
```
matthew-lidar, line 58: I think you can actually get ROS to decode a message into a custom data type to avoid this copy (probably).

matthew-lidar: https://wiki.ros.org/roscpp/Overview/MessagesSerializationAndAdaptingTypes

matthew-lidar (commented):
Looks pretty good, just a few notes. How does performance compare with the same settings?

src/os_driver_nodelet.cpp
```diff
@@ -121,11 +122,18 @@ class OusterDriver : public OusterSensor {
...
+            uint32_t max_range = pnh.param("max_range", 1000000);   // 1000m
+            auto rows_step = pnh.param("rows_step", 1);
             processors.push_back(
                 PointCloudProcessorFactory::create_point_cloud_processor(point_type, info,
                     tf_bcast.point_cloud_frame_id(), tf_bcast.apply_lidar_to_sensor_transform(),
+                    organized, destagger, min_range, max_range, rows_step,
                     [this](PointCloudProcessor_OutputType msgs) {
-                        for (size_t i = 0; i < msgs.size(); ++i) lidar_pubs[i].publish(*msgs[i]);
+                        for (size_t i = 0; i < msgs.size(); ++i)
+                        lidar_pubs[i].publish(*msgs[i]);
```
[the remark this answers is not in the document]
Samahu: I think I have already fixed this one, but without using braces.

src/point_cloud_processor.h
```diff
@@ -80,6 +83,7 @@ class PointCloudProcessor {
             auto range_channel = static_cast<sensor::ChanField>(sensor::ChanField::RANGE + i);
             auto range = lidar_scan.field<uint32_t>(range_channel);
             ouster::cartesianT(points, range, lut_direction, lut_offset,
+                               min_range_, max_range_,
```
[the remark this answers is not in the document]
Samahu: I had to duplicate this method in the driver so I would be able to needed changes with the intention to port back to the SDK at some point.

src/os_image_nodelet.cpp
Samahu: I am not familiar with this option, I need to look into it to see if it can help with the situation here

launch/sensor.launch
Samahu: 1000 is well beyond the range of any of Ouster sensors current generations so this has no limit.
Negative values would require more handling and may or may not improve understanding, in any case I would be open to changing it if you and the rest feel strongly about it.

matthew-lidar: Yeah, just ideally you would never have to update this value. If you set it to 10k maybe I'd agree, but 1000 is only like 3x our max range.

Also, if you use a key value (negative?) for disabled you can not even perform this comparison at all if the user doesn't care, saving performance.

Samahu: > Yeah, just ideally you would never have to update this value. If you set it to 10k maybe I'd agree, but 1000 is only like 3x our max range.

Surveying the net there does seem to be lidar sensor which can exceed 1k (although they are much bulkier than Ouster sensors), that being said I can increase the default max_range to 10k, not a big deal.

> Also, if you use a key value (negative?) for disabled you can not even perform this comparison at all if the user doesn't care, saving performance.

This would be implementation dependent, I mean currently I do the comparison in same pass to compute the cartesian so using negative values wouldn't help unless I bind a different operations. However, this is much more performant than the cli which needs two passes. I would rather keep as is as the gain would be minimal.

celentes (approved):

twslankard (approved):

PR #387, opened 2024-10-09
ROS-350[HUMBLE/IRON/JAZZY]: 't' timestamp field content is not plausible

### LINKED ISSUE #350: 't' timestamp field content is not plausible
**Describe the bug**
I'm switching from community driver to this driver. In my application I use the content of the `t` data field for pointcloud deskewing. The value of this field by the community driver contents the relative timeshift of each point since start of frame in nanoseconds. So for a OS1-128 in 1024x20 operation mode, the timestamp of the first and each 1024th point is zero. The ofifcial driver outputs the following timestamp values for the corresponding clouds:

`point[0] = 49416704`
`point[12] = 0`
`point[1024] = 49806592`

I don't have any idea of the meaning of these timestamps. Please let me know if you need further information.

**To Reproduce**
Configure with `native` or `original` pointcloud profile.

- OS1-128 rev. 7
- Firmware v3.1.0
- Timesyc method: PPS + $GPRMC
- ROS2 humble
- Ubuntu 22.04
- arch x86
- Release [ros2-v0.12.0](https://github.com/ouster-lidar/ouster-ros/commit/726b0d59ca61ee5a79157fac98d88832abeb7611)

Samahu:
## Related Issues & PRs
- resolves #350

## Summary of Changes
- Don't apply offset to destaggered timestamps within the point cloud
- Make sure the the timestamps align correctly when destaggerd is option is OFF

## Validation
* run the driver with default option (destaggered=ON, organized=ON)
  -  observe that the timestamps are consistent
* run the driver with default option (destaggered=OFF, organized=ON)
  - observe that the timestamps appear staggerd

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,10 @@
 Changelog
 =========
 
+[unreleased]
+============
+* [BUGFIX]: correctly align timestamps to the generated point cloud.
+
 ouster_ros v0.13.2
 ==================
 * [BUGFIX]: Make sure to initialize the sensor with launch file parameters.
--- a/ouster-ros/package.xml
+++ b/ouster-ros/package.xml
@@ -2,7 +2,7 @@
 <?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
 <package format="3">
   <name>ouster_ros</name>
-  <version>0.13.2</version>
+  <version>0.13.3</version>
   <description>Ouster ROS2 driver</description>
   <maintainer email="oss@ouster.io">ouster developers</maintainer>
   <license file="LICENSE">BSD</license>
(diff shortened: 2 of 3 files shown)


matthew-lidar (approved):

mjseppa:
This actually seems to break the timestamps within the cloud. Example from OS0 with latest FW looks bad when the lidar is in motion (spinning around Z):
image
image
Now the timestamps are just a hard cut (rviz visualization with intensity on field "t" and 0 to 99999999 interval, ~100ms).
The staggered points report the wrong timestamps. The images are with the machine spinning in place so there is difference in the positions between start and end of scan.  
Using the old 
```
            auto ts = timestamp[v_shift];
            ts = ts > scan_ts ? ts - scan_ts : 0UL;

```
the cloud timestamps are correct:
image
image

PR #406, opened 2024-11-20
ROS-396: retrieve xyz for non hit

### LINKED ISSUE #396: Unable to Retrieve X, Y, Z Coordinates and Range for Non-Hit Points with Ouster Sensor
**Is your feature request related to a problem?**

I'm currently experiencing an issue when attempting to retrieve X, Y, Z coordinates and range information for non-hit points (points where the laser did not return a measurement) using the Ouster sensor. These points are returned with (0, 0, 0) coordinates and range = 0, which makes it impossible to determine the direction of the laser beams or their intended spatial orientation. This lack of information limits our ability to apply processing and noise models to non-hit points in simulations or environmental analysis.

I would like a feature that allows the sensor to output the intended direction (in terms of azimuth and elevation) or provide a consistent directional vector for non-hit points. This could be achieved by having:

1.     The sensor provide azimuth and elevation angles for each point in the dataset, even for points with range = 0.
2.     An option to include a flag or placeholder direction vector for non-hit points in the data stream.

Having this information would enable downstream processing and analysis to interpret the spatial orientation of non-hit points without ambiguity.

I've attempted to infer the direction of non-hit points using available metadata, such as beam_azimuth_angles and beam_altitude_angles, but this approach is limited due to the lack of synchronization with each measurement and the reliance on a consistent data structure for all points. Additionally, without range data, it's challenging to ensure these inferred directions align with the actual scanning pattern.

**Targeted Platform:**

    Ouster Sensor: OS0, 128 channels.
    ROS version/distro: ROS 2 Humble.
    Operating System:  Ubuntu 20.04.
    Machine Architecture: x86_64.

Samahu:
## Related Issues & PRs
- closes #396 

## Summary of Changes
- Allow users to assign a value for no-returns (no hit)

### TODO(s):
- Expose a parameter that allows user to enable this feature and adjust the invalid (out-of-range) value.

## Validation
This image was generated by manually setting the invalid (out-of-range) value to 1m distance. 
image

### DIFF
--- a/ouster-ros/src/impl/cartesian.h
+++ b/ouster-ros/src/impl/cartesian.h
@@ -55,7 +55,9 @@ void cartesianT(PointsT<T>& points,
         const auto idx_y = col_y + i;
         const auto idx_z = col_z + i;
         if (r <= min_r || r >= max_r) {
-            pts[idx_x] = pts[idx_y] = pts[idx_z] = invalid;
+            pts[idx_x] = invalid * dir[idx_x] + ofs[idx_x];
+            pts[idx_y] = invalid * dir[idx_y] + ofs[idx_y];
+            pts[idx_z] = invalid * dir[idx_z] + ofs[idx_z];
         } else {
             pts[idx_x] = r * dir[idx_x] + ofs[idx_x];
             pts[idx_y] = r * dir[idx_y] + ofs[idx_y];
--- a/ouster-ros/src/point_cloud_processor.h
+++ b/ouster-ros/src/point_cloud_processor.h
@@ -84,7 +84,7 @@ class PointCloudProcessor {
             auto range = lidar_scan.field<uint32_t>(range_channel);
             ouster::cartesianT(points, range, lut_direction, lut_offset,
                                min_range_, max_range_,
-                               std::numeric_limits<float>::quiet_NaN());
+                               1.0f);
 
             scan_to_cloud_fn(cloud, points, scan_ts, lidar_scan,
                                         pixel_shift_by_row, i);


luis-robotic:
Hi @Samahu 

I’ve actually been working on a similar solution, but in my case, I used a VLP16 sensor, which includes a feature that allows it to send the entire point cloud. 

From what I understand, Ouster compresses the point cloud by only including return points, whereas Velodyne sends non-return points as NaN values. Aligning these non-return points with the correct ring and azimuth was somewhat challenging, so having a feature that outputs these points with their original orientation would be incredibly useful for my work.
image

Samahu:
I think there is some miscommunication, this MR still sends the entire PointCloud when you choose `organized:=true`.

The mainline of the ouster-ros driver currently sets no returns to NaN by default, these NaN values will still be sent in case of `organized` PointCloud, however, when you choose a non-organized PointCloud i.e. `organized:=false` then NaNs or no-returns values will be omitted out from the final  PointCloud.

All this MR is doing is replace NaN values with a valid point at a fixed distance which you can use to infer the direction from the generated PointCloud. So I am not sure what are you missing here.

luis-robotic:
Hi @Samahu 
I have to thank you for the proposed solution. It really covers my requirements for now. If I can contribute anything else in the future, I will be happy to do so.
Now we have a unique orientation for each point. And because we can adjust this distance, we can also filter out return and non-return points.

Maybe, as @SwiftGust suggests, you could delimit a specific field for such values, or directly enable a new field such as (is_hit = 0 or 1).

But thank you very much for the effort.

Samahu:
> I think it would be nice to set some specific value on some point field like intensity, reflectivity that notify ray did not return
> rather than multiply certain range on Cartesian coordinate and project points to Cartesian coordinate slightly further than maximum range which would allow filter-out non-returned points with range or some fields on existing codes with ease.

The issue with this proposal is that some point representation like the **PointXYZ** (which is a supported point type in the driver) doesn't have an **intensity**. Other point types have other fields but not intensity.

Regardless, you would still need to substitute zero RANGE value (no-return) with some value before computing xyz coordinates.
I think using the unit vector is mostly sufficient since no real returns would be at 1 mm distance from the sensor center, so you can use that as a threshold to detect whether this is a valid return or not.

Issue #410, opened 2024-12-11
Segmentation fault if sensor is unpowered on start-up


tombrodie:
**Describe the bug**
When the Ouster is not powered and the code is started, a segmentation fault occurs.

**To Reproduce**
1. Checkout the `master` branch on `c6f66d2c439aa27a6018a09a5e5b02fa222b52db`
2. Ensure the sensor is not on
3. Run the Ouster driver: `<include file="$(find ouster_ros)/launch/sensor.launch">`

**Debugging information**
Journal logs:
```
kernel: nodelet[1922]: segfault at 0 ip 00007f42c4649ae8 sp 00007ffe1f8638e0 error 4 in libouster_ros_nodelets.so[7f42c463a000+7a000]
```

Stack trace:
```
Core was generated by `/opt/ros/noetic/lib/nodelet/nodelet manager __name:=os_nodelet_mgr __log:=/var/'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007fac5ef63ae8 in ouster_ros::OusterSensor::stop_sensor_connection_thread() () from /opt/ros/noetic/lib//libouster_ros_nodelets.so
[Current thread is 1 (Thread 0x7fac7286d780 (LWP 68))]
(gdb) bt
#0  0x00007fac5ef63ae8 in ouster_ros::OusterSensor::stop_sensor_connection_thread() () from /opt/ros/noetic/lib//libouster_ros_nodelets.so
#1  0x00007fac5ef6482d in ouster_ros::OusterSensor::~OusterSensor() () from /opt/ros/noetic/lib//libouster_ros_nodelets.so
#2  0x00007fac5ef64e8d in ouster_ros::OusterSensor::~OusterSensor() () from /opt/ros/noetic/lib//libouster_ros_nodelets.so
#3  0x00007fac75dd4567 in void class_loader::ClassLoader::onPluginDeletion<nodelet::Nodelet>(nodelet::Nodelet*) () from /opt/ros/noetic/lib/libnodeletlib.so
#4  0x00007fac75dcb999 in boost::detail::sp_counted_impl_pd<nodelet::Nodelet*, boost::_bi::bind_t<void, boost::_mfi::mf1<void, class_loader::ClassLoader, nodelet::Nodelet*>, boost::_bi::list2<boost::_bi::value<class_loader::ClassLoader*>, boost::arg<1> > > >::dispose() ()
   from /opt/ros/noetic/lib/libnodeletlib.so
[...]

PR #411, opened 2024-12-11
Check sensor connection thread validity before joining

tombrodie:
## Related Issues & PRs
#410

## Summary of Changes
Perform a check for a nullptr on unique pointer type before usage, that otherwise has the capacity to be null when not initialised due to the sensor not being powered on start-up.

## Validation
Tested on-site


Samahu (approved):
This makes sense, it is weird that this problem was never triggered during my testing.

I think we should increment the version number in `package.xml` to `0.13.3` for tracking and optionally add the following statement to the `CHANGELOG.rst` file:
```
[BUGFIX] Segmentation fault if sensor is unpowered on start-up
```

Issue #200, opened 2023-08-28
Error code parser


TzvikiHo:
Hi,
Does the driver have an error code parser in the ros node?
I want to get these errors:
https://static.ouster.dev/sensor-docs/image_route1/image_route3/troubleshooting/troubleshooting.html

Thanks

- Ouster Sensor? OS-1
- Ouster Firmware Version? 2.4
- ROS version/distro?noetic
- Operating System? Linux

Samahu:
No, we currently don't have this feature implemented yet.

outrider-jkoch:
Just following up on this since it has been a while now. Do you think this would be implemented using `/diagnostics` or some other method?

Samahu:
Hi @outrider-jkoch, yeah it is likely that we are going to add an `alerts` topic.

PR #413, opened 2025-01-15
SW-6906: publish sensor telemetry in ouster ros

Samahu:
## Related Issues & PRs
* Related: #200

## Summary of Changes
* Define a new telemetry message and emit (it from a live sensor or bag file or pcap) 
* Always set timestamps in the PointCloud even for invalid points
* Code refactor for LidarPacketHandler

### TODO(s)
* ~Support emitting Telemetry topic in replay mode (bag|pcap)~
  - ~Probably in a separate PR~ 

## Validation
* Launch a live sensor with telemetry enabled
```bash
roslaunch ouster_ros driver.launch sensor_hostname:=... proce_mask:="...|TLM"
```
* Once sensor is connected, open a separate terminal launch, verify that the telemetry topic is up and running
```bash
rostopic echo /ouster/telemetry
```
* Verify that the telemetry messages are being sent at the same rate as the raw packets topic
```bash
rostopic hz /ouster/telemetry
```

### DIFF
--- a/include/ouster_ros/os_ros.h
+++ b/include/ouster_ros/os_ros.h
@@ -24,6 +24,7 @@
 #include <string>
 
 #include "ouster_ros/PacketMsg.h"
+#include "ouster_ros/Telemetry.h"
 #include "ouster_ros/os_point.h"
 
 namespace ouster_ros {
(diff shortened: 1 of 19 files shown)


include/ouster_ros/os_ros.h
```diff
@@ -116,6 +117,20 @@ sensor_msgs::LaserScan lidar_scan_to_laser_scan_msg(
     const uint16_t ring, const std::vector<int>& pixel_shift_by_row,
     const int return_index);
 
+
+/**
+ * Parse a LidarPacket and generate the Telemetry message
+ * @param[in] pm packet message populated by read_imu_packet
```
twslankard, line 123: Param name doesn't match. Also, should this be `read_lidar_packet`?

twslankard (approved):

msg/Telemetry.msg
```diff
@@ -0,0 +1,9 @@
+# This message defines the telemetry data for Ouster sensors.
+# Message header
+std_msgs/Header header
+# Telemetry fields for more information on these fields and their meaning, please review:
+# https://static.ouster.dev/sensor-docs/image_route1/image_route2/thermal_int_guide/therm_int_guide.html
+uint16 countdown_thermal_shutdown # the thermal shutdown countdown.
+uint16 countdown_shot_limiting    # the shot limiting countdown.
```
matthew-lidar, line 7: I know this is already merged, but why not have info about alerts in this message?

matthew-lidar: Also, ideally one would use enums here

Samahu: > Also, ideally one would use enums here

The message definition merely reflects the underlying type used by the SDK API.

matthew-lidar: I mean the alerts flags that are contained in the packet for each header. 

One should really include the enums in the message (such as how OSF has its own enums). Consumers of ROS messages shouldn't need to link to the SDK to figure out what enums mean.

Samahu: Good point, yeah we can include the raw alert flag. I can check using enums is a more viable option.

PR #419, opened 2025-01-22
ROS-416: Always copy fill all fields for organized point clouds.

### LINKED ISSUE #416: Bug in point cloud field data for structured cloud messages
### Description
In the current implementation of the ROS1 driver, the ouster_ros::Cloud<PointT> cloud object (point_cloud_processor.h:129) is reused every time to set the point cloud data before publishing. If the point cloud is "organized", the data is not cleared, further, old field data (intensity, range, signal and ambient) are not overridden when the new Cartesian coordinates are invalid, i.e., NaN (point_cloud_compose.h:140).

This has two effects that are detrimental for my current application: first, Near IR data is missing for points without range, although NIR data is, in fact, available; and second, values in the range field for the second return point cloud (second echo) are the same as the ranges from the first echo for all points without a second echo (where I would have expected simply a range of 0).

### Proposed solution
In  ouster-ros/src/point_cloud_compose.h delete lines 143 - 146 and line 148.

Samahu:
## Related Issues & PRs
- Resolves #416 

## Summary of Changes
- Always copy fill all fields for organized point clouds to have NEAR_IR channel populated regardless of having a valid range
- At the same time fill the other fields even for invalid range such that the data is overwritten.

### DIFF
--- a/src/point_cloud_compose.h
+++ b/src/point_cloud_compose.h
@@ -144,21 +144,13 @@ void scan_to_cloud_f(ouster_ros::Cloud<PointT>& cloud, PointS& staging_point,
             auto ts =
                 timestamp[ts_idx] > scan_ts ? timestamp[ts_idx] - scan_ts : 0UL;
 
-            if (xyz.isNaN().any()) {
-                if (organized) {
-                    cloud.is_dense = false;
-                    auto& pt = cloud.points[tgt_idx];
-                    pt.x = static_cast<decltype(pt.x)>(xyz(0));
-                    pt.y = static_cast<decltype(pt.y)>(xyz(1));
-                    pt.z = static_cast<decltype(pt.z)>(xyz(2));
-                    CondBinaryOp<point::has_t_v<PointT>>::run(
-                        pt, ts, [](auto& pt, const auto& ts) {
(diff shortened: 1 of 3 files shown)


src/point_cloud_compose.h
```diff
@@ -144,21 +144,13 @@ void scan_to_cloud_f(ouster_ros::Cloud<PointT>& cloud, PointS& staging_point,
...
-                    pt.y = static_cast<decltype(pt.y)>(xyz(1));
-                    pt.z = static_cast<decltype(pt.z)>(xyz(2));
-                    CondBinaryOp<point::has_t_v<PointT>>::run(
-                        pt, ts, [](auto& pt, const auto& ts) {
-                            pt.t = static_cast<uint32_t>(ts); }
-                    );
-                }
-                continue;
+            if (organized) {
+                cloud.is_dense &= xyz.isNaN().any();
```
matthew-lidar, line 148: Not a problem with your code but didnt realize that's what this message field meant officially. Seems rather pointless.

matthew-lidar (approved):

src/point_cloud_compose.h
Samahu: It is probably used by certain PCL methods.

PR #428, opened 2025-01-29
ROS-268-min-valid-cols-lidar-scan

Samahu:
## Related Issues & PRs
- Related #268

## Summary of Changes
* Skip the lidar_scan based on a predefined percentage of the minimum number of valid columns in scan

## Validation
* Launch a node as usual but enable the new `min_scan_valid_columns_ratio` parameter (set a high value)
```bash
roslaunch ouster_ros driver.launch sensor_hostname:=... min_scan_valid_columns_ratio:=0.9
```
* Observe the warning messages whenever the number of valid columns drops below the set amount:
```bash
[ WARN] [1738283732.021510354]: number of valid columns per scan 160 which is below the ratio 90%% SKIPPING SCAN
[ WARN] [1738283732.080962740]: number of valid columns per scan 304 which is below the ratio 90%% SKIPPING SCAN
```

### DIFF
--- a/src/lidar_packet_handler.h
+++ b/src/lidar_packet_handler.h
@@ -383,7 +383,7 @@ class LidarPacketHandler {
 
     int64_t ptp_utc_tai_offset_;
 
-    float min_scan_valid_columns_ratio_ = 0.7;
+    float min_scan_valid_columns_ratio_ = 0.0f;
 };
 
 }  // namespace ouster_ros
--- a/src/lidar_packet_handler.h
+++ b/src/lidar_packet_handler.h
@@ -128,6 +128,17 @@ class LidarPacketHandler {
                         *(mutexes[ring_buffer.write_head()]));
                     auto& lidar_scan = *lidar_scans[ring_buffer.write_head()];
                     result = lidar_handler(*this, pf, lidar_packet, lidar_scan);
+                    if (result) {
+                        // count the number of valid columns in the scan
+                        auto status = lidar_scan.status();
+                        size_t valid_cols = std::count_if(status.data(), status.data() + status.size(),
+                               [](const uint32_t s) { return (s & 0x01); });
+                        if (valid_cols < static_cast<size_t>(0.7 * lidar_scan.status().size())) {
+                            NODELET_WARN("valid_cols %zu, less than 70%%, SKIPPING SCAN",
+                                valid_cols);
+                            result = false;
+                        }
+                    }
                 }
                 if (result) {
                     ring_buffer.write();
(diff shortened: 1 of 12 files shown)


ovaag:
Yep, that's cleaner! Thanks for taking the time to look into this.

If we now wanted to publish this on a rostopic, I guess we could return a std::optional<size_t> from the lidar_packet_handler? Since `valid_cols` will only be relevant if result = true in lidar_packet_accumlator. 

A `/ouster/statistics` topic sounds good.

Is this something you would be interested in having upstream?

Another options is of course to have the publisher directly in the LidarPacketHandler, but that might differ a bit from how the rest of ros pubs/subs are done?

Samahu:
> If we now wanted to publish this on a rostopic, I guess we could return a std::optional<size_t> from the lidar_packet_handler? Since valid_cols will only be relevant if result = true in lidar_packet_accumlator.

I assume you are talking about publishing `valid_cols` value to a separate topic

> A /ouster/statistics topic sounds good.

If we are moving in this direction, I would say probably `/ouster/scan_stats` seems more appropriate (unless we can think of something else).

> Is this something you would be interested in having upstream?

yeah, I think it makes sense to me to add the param to filter LidarScan based on this PR, I am not sure yet about adding a scan stats or statistics topic yet.


An alternative approach would be to publish the LidarScan to a topic as is, and the let subscribed nodes do the PointCloud2 scan decoding on their own, this would be a major work that I can not take on in the near term.

src/lidar_packet_handler.h
```diff
@@ -128,6 +130,18 @@ class LidarPacketHandler {
                         *(mutexes[ring_buffer.write_head()]));
                     auto& lidar_scan = *lidar_scans[ring_buffer.write_head()];
                     result = lidar_handler(*this, pf, lidar_packet, lidar_scan);
+                    if (result) {
+                        // count the number of valid columns in the scan
+                        auto status = lidar_scan.status();
+                        size_t valid_cols = std::count_if(status.data(), status.data() + status.size(),
+                               [](const uint32_t s) { return (s & 0x01); });
+                        if (valid_cols < static_cast<size_t>(min_scan_valid_columns_ratio_ * lidar_scan.status().size())) {
+                            NODELET_WARN_STREAM("number of valid columns per scan " << valid_cols
```
matthew-lidar, line 139: Might be worth saying vs the expected number of columns e.g. 890/1024. Also the %% seems odd to me. Perhaps you were trying to escape as if this was printf, but it isnt necessary as both end up in the log.

matthew-lidar (approved):
One small thing about the warning print, otherwise looks good!

src/lidar_packet_handler.h
Samahu: make sense

Samahu: yeah I intially I started with NODELET_WRAN which needed the %% but now that I switchoed NODELET_WARN_STREAM i should probably use \%

PR #436, opened 2025-02-16
Reduce sensor_msgs::msg::PointCloud2 size by 50%, when using pcl::PointXYZI

doisyg:
This should fix the size of the message size, that is currently 32 byte per point, instead of 16 bytes, because of padding.

Original work from @facontidavide

### DIFF
--- a/ouster-ros/src/point_cloud_processor.h
+++ b/ouster-ros/src/point_cloud_processor.h
@@ -18,6 +18,8 @@
 #include "lidar_packet_handler.h"
 #include "impl/cartesian.h"
 
+#include <sensor_msgs/point_cloud2_iterator.hpp>
+
 namespace ouster_ros {
 
 // Moved out of PointCloudProcessor to avoid type templatization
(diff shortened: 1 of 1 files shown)


Samahu:
Hi @doisyg, this seems interesting but I am wondering rather than changing the point cloud content on the fly, I think it would be easier and much more efficient to define a new point type that doesn't employ padding. For example following the example provided in PR #290 that would be something around the line:

```
struct EIGEN_ALIGN16 _Ouster_PointXYZI {
     union EIGEN_ALIGN16 {
          float data[4];
          struct {
            float x;
            float y;
            float z;
            uint32_t intensity;
          };
     }; 
}; 
``` 

Then ouster-ros driver will handle pretty much the rest.

Samahu:
@doisyg I have made a quick draft of mt proposal to reduce the point_size directly through the ouster_ros driver. This would rather be a different point_type but it is has the same fields as `pcl::PointXYZI`. Let me know if this seems better approach.

Samahu:
absolutely!

Samahu:
Here is the ROS2 PR: #439

doisyg:
Nice, thanks a lot, tested and working, closing this!

PR #456, opened 2025-05-06
Update README.md

akshaygulabrao:
typo in sourcing path for installing the ouster_ros driver


Samahu:
The reason you had the install folder at `ros2_ws/src/install/setup.bash` instead of `ros2_ws/install/setup.bash` is because you probably invoked the `colcon build` inside the `ros2_ws/src` folder, which isn't the proper place to invoke this command in, unlike ros1 catkin build, the ros2 colcon build system will create the `build`, `install` and `log` folders inside the current working folder.

PR #460, opened 2025-05-28
PointCloud mask

Samahu:
## Related Issues & PRs
- None

## Summary of Changes
- Implemented a mask filter on the received LidarScan.
- The mask is applied to the cloud + the image messages
- If the supplied mask size is different from the received LidarScan, the driver will resize the mask to match incoming scans size.

## Validation
* Test out a sequence without applying a mask:
image
* Generate any mask composed of Black/White pixels, such as:
image
* Apply the mask as follows:
```bash
roslaunch ouster_ros driver.launch sensor_hostname:=<source_url> mask_path:=<path/to/mask.png>
```
* Observe the masked image in RVIZ/image_view:
image
* Observe the mask cloud in RVIZ/cloud_view
image

### DIFF
--- a/src/image_processor.h
+++ b/src/image_processor.h
@@ -15,6 +15,8 @@
 // clang-format on
 
 #include <sensor_msgs/image_encodings.h>
+#include <opencv2/opencv.hpp>
+#include <opencv2/core/eigen.hpp>
 
 #include "ouster/image_processing.h"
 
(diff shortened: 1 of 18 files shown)


src/point_cloud_processor.h
```diff
@@ -82,7 +98,8 @@ class PointCloudProcessor {
         for (int i = 0; i < static_cast<int>(pc_msgs.size()); ++i) {
             auto range_ch = static_cast<sensor::ChanField>(sensor::ChanField::RANGE + i);
             auto range = lidar_scan.field<uint32_t>(range_ch);
-            ouster::cartesianT(points, range, lut_direction, lut_offset,
+            auto range_masked = mask.rows() == range.rows() && mask.cols() == range.cols() ? range * mask : range;
```
twslankard, line 101: Should there be a way to let the user know they specified an incorrectly-sized image? Otherwise, they might be confused as to why the mask they have is ineffective.

twslankard (approved):

src/point_cloud_processor.h
Samahu: Added a condition that checks if mask match the sensor resolution

src/image_processor.h
```diff
@@ -173,19 +165,48 @@ class ImageProcessor {
...
+            image.cols != static_cast<int>(width)) {
+            std::stringstream ss;
+            ss << "Mask image size (" << image.rows << "x" << image.cols
+               << ") does not match the expected dimensions ("
+               << height << "x" << width << ").";
+            throw std::runtime_error(ss.str());
+        }
+        Eigen::MatrixXi eigen_img(image.rows, image.cols);
+        cv::cv2eigen(image, eigen_img);
+        mask = eigen_img.cast<pixel_type>() / 255;
```
matthew-lidar, line 207: What if someone uses a 16 bit image or accidentally doesnt quite make the pixels 255? We should just set the mask to exactly 1 if the pixel isnt black to avoid ways to mess this up

src/point_cloud_processor.h
```diff
@@ -66,6 +71,31 @@ class PointCloudProcessor {
...
+            image.cols != static_cast<int>(width)) {
+            std::stringstream ss;
+            ss << "Mask image size (" << image.rows << "x" << image.cols
+               << ") does not match the expected dimensions ("
+               << height << "x" << width << ").";
+            throw std::runtime_error(ss.str());
+        }
+        Eigen::MatrixXi eigen_img(image.rows, image.cols);
+        cv::cv2eigen(image, eigen_img);
+        mask = eigen_img.cast<uint32_t>() / 255;
```
matthew-lidar, line 98: This code is completely duplicated. It should be moved to a common file.

Samahu: I was just making this change.

src/point_cloud_processor_factory.h
```diff
@@ -144,37 +145,38 @@ class PointCloudProcessorFactory {
...
                         organized, destagger, min_range, max_range, rows_step,
-                        post_processing_fn);
+                        mask_path, post_processing_fn);
                 case UDPProfileLidar::PROFILE_RNG19_RFL8_SIG16_NIR16_DUAL:
                     return make_point_cloud_processor<
                         Point_RNG19_RFL8_SIG16_NIR16_DUAL>(
                         info, frame, apply_lidar_to_sensor_transform,
                         organized, destagger, min_range, max_range, rows_step,
-                        post_processing_fn);
+                        mask_path, post_processing_fn);
```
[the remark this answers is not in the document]
Samahu: Totally, just when the right time that is. I think I have a TODO in the ROS2 branch to refactor this when I started adding these new filters.

src/image_processor.h
matthew-lidar: I'm still concerned about images that aren't quite white. Is there any reason counting any point that isnt exactly 0 as white is a problem?

Samahu: The input image is a mask, ideally the user would pass a binary image. If I pick a given threshold let's say the middle of 255 then users may need to know this middle point. Setting the expectation 255 as a pass and anything else rejected would be easier to communicate.

Samahu: flipped

PR #467, opened 2025-06-26
Fix .is_dense logic for pointcloud

ovaag:
## Summary of Changes

With the old logic, cloud.is_dense will only be set to true if every single point in the cloud has a NaN coordinate.
Replace
`cloud.is_dense &= xyz.isNaN().any();`
with
`cloud.is_dense &= !xyz.isNaN().any();`

now is_dense is true if no points in cloud are nan, false otherwise

## Validation
Testing by altering the `invalid` arg in ouster::cartesianT

Before:
* `std::numeric_limits<float>::quiet_NaN()` -> `is_dense = False`
* `0.0f` -> `is_dense = False` 

After:
* `std::numeric_limits<float>::quiet_NaN()` -> `is_dense = False`
* `0.0f` -> `is_dense = True`

### DIFF
--- a/src/point_cloud_compose.h
+++ b/src/point_cloud_compose.h
@@ -145,7 +145,8 @@ void scan_to_cloud_f(ouster_ros::Cloud<PointT>& cloud, PointS& staging_point,
                 timestamp[ts_idx] > scan_ts ? timestamp[ts_idx] - scan_ts : 0UL;
 
             if (organized) {
-                cloud.is_dense &= xyz.isNaN().any();
+                // false if any point in cloud has NaN values
+                cloud.is_dense &= !xyz.isNaN().any();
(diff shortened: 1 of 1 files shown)


Samahu (commented):
Yes, you are right .. I probably missed adding the ! when implementing it.
Good catch!

Needed changes before merge:
- [ ] Mention the fix CHANGELOG.txt
- [ ] Increment version minor in `package.xml`

Otherwise, I can merge your changes into a temp branch then implement these changes myself before merging into **master**.

Thanks!

Samahu:
Merged into mainline for ROS1 & ROS2

ovaag:
Thanks! Been away on holiday until now :slightly_smiling_face:

PR #477, opened 2025-07-28
Ros2 fix opencv dependency

Erumeldor:
## Summary of Changes
Add OpenCv as library dependency

## Validation
Use os_ros.h in another node - without dependency it is missing the "opencv2/opencv.hpp"


ouster-ros/CMakeLists.txt
```diff
@@ -77,6 +77,7 @@ set(ouster_ros_library_deps
   pcl_conversions
   tf2
   tf2_eigen
+  OpenCV
```
Samahu, line 80: I am not sure this is required since we explicity link OpenCV for the **ouster_ros_library** target

Erumeldor: Well, I am totally not confident in CMake files, just the information above.
We switched from an old version of ouster-ros to the newest, we are importing the os_ros.h, and it failed because of not finding "opencv2/opencv.hpp".
Adding this line solved it.

PR #491, opened 2025-09-16
Fix sensor healthy

ralwing:
## Summary of Changes

A few small fixes:
 * Removed the sensor connection status : which in fact reports whether the connection thread works.
 * Fixed the is_healthy_function - now reporting if the packets are received.

## Validation

Screenshot from 2025-09-16 11-58-59

Warning when there's no connection

Screenshot from 2025-09-16 11-59-17


ralwing:
invalid repo selected.

PR #526, opened 2026-03-13
os_sensor_nodelet: Fix sensor fw version check

ovaag:
Just saw a version warning from this. 
'fw_rev' just gives a hash, while 'image_rev' has the version string expected by ouster::sdk::core::version_from_string

https://github.com/ouster-lidar/ouster-ros/blob/24f89d10425edd0b6422397ad4bb734810e63648/src/os_sensor_nodelet_base.cpp#L50 uses image_rev, so I suppose it was just forgotten in os_sensor_nodelet.


Samahu:
Good catch! .. yes.. `image_rev` is the official way to check for the FW version.

Samahu (approved):

PR #534, opened 2026-04-08
Qos override

doisyg:
## Summary of Changes

Allow QoS override using standard ros2 mechanism

## Validation

Tested on Dexory stack


doisyg:
@Samahu any opinion on this one ?

Issue #544, opened 2026-05-13
Refactor Point Cloud Masking Feature as it is Currently Unintuitive


PhilippSchmaelzle:
Hey, 
I am a active user of your driver and just now migrated to the latest version.
So now I tried to make use of the *masking* feature via the `mask_path` parameter.

## Setup and Problem
I manually created a mask with *ImageJ* by loading a `near_ir` image and drawing the mask on top of it. 
Maybe not the most professional way but for sure a widely used one 😄 
My goal is to mask out all points which would hit the ego vehicle - the sensor is mounted on the front right corner of the roof. Tilted by 45° and only 180° FoV.

<img width="1024" height="128" alt="near_ir image" src="https://github.com/user-attachments/assets/7d2ae793-b0c7-4f75-8a90-879f64618371" />

The resulting mask would be this.

<img width="1024" height="128" alt="mask image" src="https://github.com/user-attachments/assets/fad8fed6-d2ce-48a1-b352-ec1e70b01bcc" />

Without any masking, the point cloud would look like this:

<img width="1492" height="1009" alt="Point Cloud without masking" src="https://github.com/user-attachments/assets/cbd8dca0-cd54-45fe-adc5-23fadcba8de0" />

With your original `load_mask` implementation, the resulting point cloud looks like this:

<img width="1492" height="1009" alt="Point Cloud with original load_mask code" src="https://github.com/user-attachments/assets/d680443f-c552-45de-b2d7-4edab7c03149" />

## Proposal

I personally think it is more intuitive if the mask is *human readable*. 
But as the masking is applied on the yet still staggered point cloud the mask would need to be staggered as well.
Therefor I propose, that the mask will be staggered within the `load_mask` method.

With a proposed refactoring the result would look like this:

<img width="1492" height="1009" alt="Point Cloud with patched load_mask code" src="https://github.com/user-attachments/assets/08e77cdc-7552-4cbc-80dd-ac8caca4cfe9" />

and the *image path*:

<img width="1024" height="128" alt="Image" src="https://github.com/user-attachments/assets/6aa0220f-be00-4fbb-9bd4-8d7fdd09a159" />

## Bug

While writing this, I checked the the *image path* as well. 
There you already work on destaggered data.
So there is actually a bug in your current system where the same mask would result in different outputs on the *image path* and *point cloud path*.

Therefor I applied the staggering in my PR only on the *point cloud path*.
But you could also think about a solution, where you apply the masking **after** you have destaggered the point cloud.

PR #546, opened 2026-05-13
Patch apply staggering to mask

PhilippSchmaelzle:
## Related Issues & PRs
- https://github.com/ouster-lidar/ouster-ros/issues/544

## Summary of Changes
Masking of the point cloud is applied before the data is destaggered.
Therefor the required mask needs to be staggered -> which is hard to provide when creating the mask manually.

The change applies a staggering to the loaded (destaggered, human readable) mask.

But this only applies for the path of the 3D point cloud.
As the masking on the image path is done after destaggering.
So this addresses a active bug in the driver as well.

## Validation
I validated it on an OS0 Rev7 with FW3.2 and the latest `ros2` branch.

Validation was done visually on the image and point cloud path.
But, only with `destagger==true` parameter (default).


Samahu:
This makes sense .. I think this is what I have done inside the main sdk .. will take a look and try to merge it soon

PR #549, opened 2026-05-18
[ROS2] Harden timing, packet processing, lifecycle, and PCAP replay

jcfurey:
## Summary

This is the runtime-correctness and performance slice of the work developed and
exercised on the `cam_wip` integration branch, rebased onto the current `ros2`
branch. It deliberately preserves ROS time, simulated time, and bag playback
rate semantics: the live ROS processing path does not introduce wall-clock
throttling or silently decimate scans.

### Timing and scan correctness

- Make scan-timestamp interpolation epoch-safe and exact with integer
  arithmetic, checked offsets, and overflow-safe fallbacks.
- Derive scan spacing from the sensor `DataFormat`, rather than an assumed
  frequency.
- Store timestamps per ring-buffer slot under the same slot lock used for the
  scan, eliminating the timestamp/scan race.
- Drop all-zero-timestamp scans, reset the estimator after the discontinuity,
  and keep ROS-time anchoring correct across the drop.
- Accept missing scan fields and fix point-cloud field/cursor shifting bugs,
  with regression coverage.

### Packet path, concurrency, and performance

- Replace timeout polling with predicate-based wakeups that cannot lose a
  notification; start workers only after construction completes and make
  shutdown deterministic.
- Never silently decimate when consumers fall behind. A full ring produces an
  explicit drop instead, preserving the timing of every scan that is published.
- Reuse packet buffers, compute RGB only when a subscriber/output consumes it,
  and expose standard point-cloud QoS overrides.
- Snapshot lock-free ring indices once per operation and keep slot ownership
  consistent.

### Lifecycle, replay, and configuration

- Harden metadata parsing, reconnect behavior, lifecycle teardown, and
  publisher cleanup.
- Validate truncated PCAP packets, pace replay from a steady clock, make sleeps
  interruptible, avoid an empty-PCAP loop spin, and do not convert lifecycle
  deactivation into global ROS shutdown.
- Fix launch/config blockers, including bag filename selection and the
  `laser_frame`/`lidar_frame` mismatch.

## Validation

- Rebased on current `ouster-lidar/ouster-ros:ros2`.
- Combined with #553 (which repairs test discovery): clean ROS 2 Lyrical build;
  23 GoogleTests and 24 total test-result entries passed.
- Full #549 + #552 + #553 composition: 25 GoogleTests and 26 total test-result
  entries passed.
- PCAP smoke tests covered normal recorded pacing and empty-file looping; the
  empty loop exits without busy-spinning.

## Relationship to the other PRs

- #553 is the independent build/test modernization slice. It is only used above
  to ensure this PR's tests are actually compiled and run on Lyrical.
- #552 is the camera and pinhole slice and can be reviewed separately.

### DIFF
--- a/ouster-ros/CMakeLists.txt
+++ b/ouster-ros/CMakeLists.txt
@@ -233,6 +233,7 @@ if(BUILD_TESTING)
   ament_add_gtest(${PROJECT_NAME}_test
     src/os_ros.cpp
     test/test_main.cpp
+    test/lidar_packet_handler_test.cpp
     test/lock_free_ring_buffer_test.cpp
     test/point_accessor_test.cpp
     test/point_transform_test.cpp
--- a/ouster-ros/config/community_driver_config.yaml
+++ b/ouster-ros/config/community_driver_config.yaml
@@ -8,7 +8,7 @@ ouster_driver:
     imu_port: 7503
     lidar_port: 7502
     sensor_frame: laser_sensor_frame
-    laser_frame: laser_data_frame
+    lidar_frame: laser_data_frame
     imu_frame: imu_data_frame
 
     # if False, data are published with sensor data QoS. This is preferrable
--- a/ouster-ros/config/driver_params.yaml
+++ b/ouster-ros/config/driver_params.yaml
@@ -103,6 +103,16 @@ ouster/os_driver:
     # data QoS. This is preferrable for production but default QoS is needed for
     # rosbag. See: https://github.com/ros2/rosbag2/issues/125
     use_system_default_qos: false
+    # PointCloud2 publisher QoS may be overridden independently using ROS 2's
+    # standard qos_overrides parameters. The topic key is fully resolved, so
+    # update it if the default `ouster` namespace changes; repeat for points2
+    # with dual returns. Example:
+    # qos_overrides:
+    #   /ouster/points:
+    #     publisher:
+    #       reliability: reliable
+    #       history: keep_last
+    #       depth: 10
     # point_type[optional]: choose from: {original, native, xyz, xyzi, o_xyzi,
     #                                     xyzir, xyzrgb, xyzrgba, color_point}
     # Here is a breif description of each option:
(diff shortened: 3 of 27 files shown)


jcfurey:
I have humble, jazzy, kilted, and lyrical all green on the built-in CI across the three RMWs. Rolling is still red for all three RMW jobs. Rosdep keys likely need the Ubuntu resolute image to resolve correctly. That was the reason for the post da75ef0 force pushes. I did try some dockerfile workarounds, but it was growing beyond the scope of this PR.

jcfurey:
I also have a cardinal-direction pinhole panel publisher (default front/left/rear/right at lidar_frame yaws 0/90/180/270°, this should be tweaked based on lidar yaw placement) that I'd like to file as a follow-up after #549 lands. It depends on #549's ouster-sdk v0.16.2 bump for the ouster::sdk::core::image::AutoExposure namespace and is otherwise additive. It should rebase cleanly onto #549 + #550.

I also test-merged ros2-support-native-rgb-lidar on top of #549 + #550 locally — integrates cleanly with 4 small conflicts, all on the SDK-adapter sites this PR modernizes (LidarMode::MODE_UNSPEC → struct sentinel, lidar_mode_of_string optional). The RGB branch also bumps ouster-sdk to a SHA 3 commits past v0.16.2, so merging it after #549 is a trivial fast-forward; merging before would make this PR's SDK bump look like a downgrade.

jcfurey:
Rebased on top of #547 and trimmed this down to just the runtime bug fixes: the scan-timestamp race, no longer publishing scans with a zero timestamp, properly rejecting bad lidar_mode values (the current error handling never actually triggers), and a few launch/param fixes. Dropped the SDK bump since #547 already covers it, and I’ll send the CMake/packaging cleanup separately once this is in. Down to 5 commits, no submodule changes, merges cleanly.

jcfurey:
Status + a couple of corrections to earlier notes in this thread:

- Rebased on #547 and trimmed to runtime fixes only — 5 commits, no submodule
  change, merges cleanly onto `ros2`. Ready for review.
- Correction: the `ouster::sdk::core::image::AutoExposure` namespace move comes
  from **#547**, not an SDK bump in this PR (that bump was dropped in the trim).
- The camera_info + cardinal-pinhole follow-up is now **#552** (supersedes the
  closed #550).
- Easy to miss in the diff: `OusterSensor::cleanup()` now cancels and drops
  `reconnect_timer` — its lambda captures `this`, so a pending reconnect could
  otherwise fire after teardown (use-after-free) on reconfigure. It rides along
  in the "minor cleanups" commit but it's a real correctness fix.
- CI was green on humble / jazzy / kilted / lyrical × 3 RMW on the pre-rebase
  push; the current head just needs a maintainer to approve the workflow run.
  Happy to fold in any review feedback.

jcfurey:
I'm probably going to redo this one...

PR #551, opened 2026-05-31
add new launch parameter no_return_is_inf to the os_driver node

Revilum:
## Related Issues & PRs
No related issues or PRs exist. I decided to not open a feature request for a feature I was going to implement anyway :P

## Summary of Changes
Added a new launch parameter `no_return_is_inf` to the `os_driver` node. If set to true, the distance of a laser scan with no reflection will be set to `inf` instead of `0` in the ROS `LaserScan` message.

While I could not find a REP which defines `inf` values in a `LaserScan` message, a lot of existing ROS systems expect in a no reflection case an `inf` value instead of `0`. An example would be the [obstacle layer](https://docs.nav2.org/configuration/packages/costmap-plugins/obstacle.html#obstacle-layer-parameters) for Costmaps in nav2.

## Validation
The code with the new parameter has been tested on a OS0 LIDAR. I don't have access to other ouster hardware which I could use to test this new feature.


gstorer:
For reference the this is the REP: https://www.ros.org/reps/rep-0117.html

Also included is:
-inf: to close
nan: invalid measurement
+inf: no return

We would also prefer the Ouster driver conformed to this standard.

Samahu:
This makes sense to me ..  we do something similar in the pointcloud processing where we convert 0 or out of range value into `nans` https://github.com/ouster-lidar/ouster-ros/blob/d72664fe4963ed5da2b37128abdedfd77e2d2137/ouster-ros/src/point_cloud_processor.h#L95

I'd say if this what the REP requires then we probably don't to add a parameter for it.

Revilum:
> This makes sense to me ..  we do something similar in the pointcloud processing where we convert 0 or out of range value into `nans` https://github.com/ouster-lidar/ouster-ros/blob/d72664fe4963ed5da2b37128abdedfd77e2d2137/ouster-ros/src/point_cloud_processor.h#L95
> 
> I'd say if this what the REP requires then we probably don't to add a parameter for it.

Ok I will remove the the parameter later and rewrite every `0` measurement as `+inf`. I am not that experienced with the ouster SDK to know wether it information about "too close" and erroneous measurements or is it ok if we keep this PR to only forfill the `+inf` standard?

Samahu:
Unfortunately, the sensor itself makes no distinction between too close or no return when reporting the value.

Revilum:
I commited the requested changes, but I haven't been able to test it again

Revilum:
works now :+1:

ouster-ros/src/os_ros.cpp
```diff
@@ -245,7 +245,13 @@ sensor_msgs::msg::LaserScan lidar_scan_to_laser_scan_msg(
         auto v_shift = (v + ls.w - pixel_shift_by_row[u] + ls.w / 2) % ls.w;
         auto src_idx = u * ls.w + v_shift;
         auto tgt_idx = ls.w - 1 - v;
-        msg.ranges[tgt_idx] = rg[src_idx] * ouster::sdk::core::RANGE_UNIT;
+
+        float r = rg[src_idx] * ouster::sdk::core::RANGE_UNIT;
+        if (rg[src_idx] == 0) {
+            r = std::numeric_limits<float>::infinity();
+        }
+        msg.ranges[tgt_idx] = r;
```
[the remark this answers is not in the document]
Samahu: Disregard this.

ouster-ros/src/os_ros.cpp
```diff
@@ -245,7 +245,13 @@ sensor_msgs::msg::LaserScan lidar_scan_to_laser_scan_msg(
         auto v_shift = (v + ls.w - pixel_shift_by_row[u] + ls.w / 2) % ls.w;
         auto src_idx = u * ls.w + v_shift;
         auto tgt_idx = ls.w - 1 - v;
-        msg.ranges[tgt_idx] = rg[src_idx] * ouster::sdk::core::RANGE_UNIT;
+
+        float r = rg[src_idx] * ouster::sdk::core::RANGE_UNIT;
+        if (rg[src_idx] == 0) {
+            r = std::numeric_limits<float>::infinity();
```
Samahu, line 251: I would actually use nan, instead because based on the sensor docs we use 0 to indicate no valid measurement:
```suggestion
            r = std::numeric_limits<float>::quiet_NaN());
```

Of course we need to test this works fine with RVIZ similar to the pointcloud and include a [BREAKING] statement in the CHANGELOG.txt.

Let me know if you want to do these or I could help with that.

Revilum: I can do that but then I would re-add the option to change the `NaN` to `+Inf` so that the user can decide whether `0` are invalid or out-of-range measurements. In my case, I care a lot about `+Inf` measurements because I operate the LIDAR in a very open environment.

Samahu: I am okay with that. Let's put back the option and let user choose how to interpret the `0` measurement from Ouster and sorry for the confusion but I wanted to minimize the number of settings that the user need to set.

Regarding the default value of this param:
* If we set to `0` (although it is wrong) there would be no breaking change 
* If we set to `nan` it would be inline with how we handle the default the pointcloud but then we need to label as a BREAKING change/behavior.

I am fine with either.

Revilum:
okay I reverted the PR to include the parameter again and set the default value to `NaN`. Also tested it and it works :+1:

Samahu (approved):
looks good

PR #558, opened 2026-07-18
[ROS2] Fix cropped scan when using both mask_path + v_reduction

nhewitt99:
## Related Issues & PRs
#462 introduced this bug when feature was implemented
#546 also modifies the image mask feature but does not address this fix

## Summary of Changes
Just fixing a super quick bug I encountered: when using both v_reduction and a mask, the mask is just applied to the first `height / v_reduction` rows, and the remaining fraction of the cloud is dropped. 

This is because the PointCloudProcessor resizes the mask to `pixels_per_column / row_step` height, but the mask is then applied to the whole (unreduced) scan data in the `scan_to_cloud_fn`, and then the reduction is finally applied.

This change fixes that, and also nitpicks the mask resize warning to report width x height as expected.

## Validation
With a fully white 1024x128 image as the mask and v_reduction set to 4, we now properly see the whole scan region. (below: before, after)
cloud-reduced-before
cloud-reduced-after

I also verified that more interesting masks (not fully white pixels) work as expected with v_reduction set to either 1 or 4.

Testing was performed on an OS0 128 rev 7, firmware 3.2.0.

### DIFF
--- a/ouster-ros/src/os_ros.cpp
+++ b/ouster-ros/src/os_ros.cpp
@@ -182,7 +182,7 @@ void warn_mask_resized(int image_cols, int image_rows,
                        int scan_height, int scan_width) {
     auto logger = rclcpp::get_logger("ouster_ros");
     RCLCPP_WARN_STREAM(logger, "Mask image has size (" << image_cols << "x" << image_rows << ")"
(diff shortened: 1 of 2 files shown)


Samahu (approved):
This looks good to me

PR #559, opened 2026-07-20
Cast packet buffers before swapping

Tobias-Fischer:
This is part of an effort to contribute RoboStack downstream patches back upstream.

Origin: RoboStack patch/ros-rolling-ouster-ros.patch, authored by Daisuke Nishimatsu.

The generated message buffer type may not expose std::vector swap directly on all supported builds. Casting the message buffer to its vector base before swapping keeps the existing transfer pattern while making the call portable.

I intentionally left out the CMake target-link modernization from the RoboStack patch because it overlaps existing upstream PR #553. The version.h major/minor macro part lives in the Ouster SDK submodule rather than in this ROS package source tree.

### DIFF
--- a/ouster-ros/src/os_sensor_node.cpp
+++ b/ouster-ros/src/os_sensor_node.cpp
@@ -1179,12 +1179,14 @@ void OusterSensor::stop_sensor_connection_thread() {
 }
 
 void OusterSensor::on_lidar_packet_msg(const LidarPacket&) {
-    lidar_packet_msg.buf.swap(lidar_packet.buf);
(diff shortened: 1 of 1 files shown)


Tobias-Fischer:
FYI @wep21 @traversaro, this was opened from the RoboStack upstreaming tracker: https://github.com/RoboStack/robostack.github.io/issues/16

Source patch: https://github.com/RoboStack/ros-rolling/blob/main/patch/ros-rolling-ouster-ros.patch

Samahu (approved):

PR #552, opened 2026-06-16
[ROS2] Add CameraInfo, optical frames, and cardinal pinhole views

jcfurey:
## Status / dependency

This PR remains closed/parked while #553 is reviewed. Its branch has been
rebuilt as a clean linear stack directly on the current #553 head (`baa2c3c`),
so it no longer contains the old merge commit or duplicate copies of the #553
history.

The stack intentionally excludes the separate packet-field experiment and does
not add or change a license file.

### `os_image` CameraInfo and optical frame

- Add opt-in panorama `CameraInfo` publication (`publish_camera_info` defaults
  to `false`) with corrected horizontal/vertical intrinsics and a destagger-aware
  `column_window` ROI.
- Warn clearly that the equirectangular panorama is not a true pinhole camera.
- Add an optional REP-103 optical frame and static transform shared by images
  and CameraInfo.

### Rectified cardinal pinhole views

- Add `os_pinhole`, which resamples the panorama into configurable rectified
  pinhole panels (front/left/rear/right by default), each with matching
  `CameraInfo` and optical-frame semantics.
- Validate projection parameters, bound auto-derived output height before
  allocation, wrap rounded columns at the panorama seam, and handle dual-return
  NIR correctly.
- Add focused regression tests for seam wrapping, allocation bounds, and the
  cardinal panels' center rays.

### Playback, lifecycle, and portability hardening

- Allow `OusterCloud`, `OusterImage`, and `OusterPinhole` to initialize from a
  metadata file for packet-only bag playback workflows, while ignoring an
  unchanged copy later delivered on the latched metadata topic.
- Reuse packet buffers, avoid RGB work unless consumed, reject undersized data,
  and make image/pinhole metadata reconfiguration and teardown safe.
- Construct static TF broadcasters through the interface-based API on
  `tf2_ros >= 0.44`, while retaining the older ROS 2 API fallback.
- Plumb the new `os_image` options through the composite, PCAP, and bag replay
  launches.

## Validation

- Exact rebuilt branch on #553 (`baa2c3c`), ROS 2 Lyrical / `tf2_ros 0.45.9`,
  Release, `BUILD_TESTING=ON`, `BUILD_PCAP=ON`: clean build.
- 21 GoogleTests across five suites passed; `colcon test-result` reports 22
  tests, 0 errors, 0 failures, and 0 skipped.
- Standalone `OusterCloud`, `OusterImage` (CameraInfo + optical frame enabled),
[...]

PR #564, opened 2026-08-26
[ROS2] Add CameraInfo, optical frames, and cardinal pinhole views

jcfurey:
ROS 2 implementation of #441. Supersedes #552.

#553 has merged, and this branch is rebased onto `ros2`.

## Summary

- Add optional `CameraInfo` and optical frames to `os_image`.
- Add `os_pinhole` for configurable rectilinear panels with matching
  `CameraInfo`, optical frames, and `32FC1` depth.
- Support automatic cropping, explicit ROIs, configurable panel geometry, and
  independent output selection.
- Only advertise channels available in the active UDP profile. REV8 RGB remains
  opt-in.
- Include four cardinal defaults, an OSDome preset with a top panel, and the
  related replay and recording wiring.
- Rebuild the processing pipeline when sensor metadata changes.

`os_image` remains equirectangular, so its `CameraInfo` is disabled by default
and should only be used as a display approximation. Use `os_pinhole` for
projection or back-projection.

## Geometry notes

`CameraInfo` describes the full configured panel, while `CameraInfo.roi`
describes the emitted crop. Unsupported depth samples are `NaN`.

Depth uses the nearest calibrated lidar sample. Back-projection preserves that
sample's optical-axis depth, but X/Y lie on the output pixel-centre ray, so the
resulting cloud is not an exact copy of the raw XYZ cloud.

With `parent_frame: os_lidar`, yaw 0 points toward the external connector, not
the housing front. Vehicle-relative panel names must account for the sensor
mounting.

## Validation

- 54 gtests and three ROS launch-test assertions passed.
- Tests cover all supported lidar profiles, 16 through 256 beams, and widths
  from 512 through 4096 columns.
- Real OS0, OS1, OS2, single-return, dual-return, and REV8 data were exercised.
- An OS1-64 firmware 3.0.1 bag passed ROS/RViz testing at 1024x10, including all
  four panels and 100 synchronized depth/`CameraInfo` back-projections.
- The whole workspace passed 1,247 tests.

Raw packet recordings were unavailable for RGB, OSDome, and five uncommon
profiles; those paths have metadata-driven or synthetic coverage.


jcfurey:
I have rebased in line with the as-now current ros2 branch and my #553 PR. Will need to rebase again more than likely

jcfurey:
I likely have a couple of more commits to land. After those, I'll be happy again with this PR.

Issue #303, opened 2024-02-23
How do I change the QOS in ROS2?


tom-bu:
**Describe your question**
Is there a good way to change the QOS to be best effort and a history of 1? Currently, it seems it's only editable by  changing the use_system_default_qos parameter, which I'm also not sure where it is being called. When use_system_default_qos is false, the QOS is best effort and history of 5 but a long history with point clouds may be consuming too much memory. When it's set to true, it's set to reliable and a history of 1 but reliable is not appropriate for sensor data. Currently, I have been inspecting the QOS by doing ros2 topic info -v /ouster/points. Is there a QOS parameter file that the driver loads to modify the publisher?

**Platform**
- OS-1
- ROS2 Humble
- Linux
- x64

yashmewada9618:
Hi @tom-bu,

The `use_system_default_qos` param is been called [here](https://github.com/ouster-lidar/ouster-ros/blob/3f9790713efb56099a7348aec6bdefa3e6807279/ouster-ros/src/os_cloud_node.cpp#L74). 

You can create your own rmw QoS policy like this:

```c++
static const rmw_qos_profile_t qos = {RMW_QOS_POLICY_HISTORY_KEEP_LAST,
                                          1, // depth
                                          RMW_QOS_POLICY_RELIABILITY_RELIABLE,
                                          RMW_QOS_POLICY_DURABILITY_VOLATILE,
                                          RMW_QOS_DEADLINE_DEFAULT,
                                          RMW_QOS_LIFESPAN_DEFAULT,
                                          RMW_QOS_POLICY_LIVELINESS_SYSTEM_DEFAULT,
                                          RMW_QOS_LIVELINESS_LEASE_DURATION_DEFAULT,
                                          false}; // RMW QOS policy for Ouster lidar, without this the sync won't work
auto lidar_qos = rclcpp::QoS(rclcpp::QoSInitialization::from_rmw(my_custom_qos_profile), my_custom_qos_profile);
```

Hope this helps.

Samahu:
> Is there a QOS parameter file that the driver loads to modify the publisher?

The only way you can control the QoS currently would be through the `use_system_default_qos` which can be set through the `driver_params.yaml`, which allow you to switch between reliable and sensor data. If you need more fine tune control then you need to follow the advice from @yashmewada9618 which requires you to modify the driver code.

PR #566, opened 2026-09-01
Override system-default-qos with packet-qos

Samahu:
## Related Issues & PRs
- Related #303

## Summary of Changes
* Override the QoS Depth for packet topic whether when using System Defaults (reliable) or when not using System Defaults (best effort)

### Notes to reviewers
* A better approach would be to use something similar to `driver.launch.xml` in which packets are parsed in the same node that produces them .. but this is an easier path to apply the fix than a whole refactor at this moment.

## Validation
Download a bag file and replay it via the command:

```bash
ros2 launch ouster_ros replay.launch.xml bag_file:=input.bag
```

### DIFF
--- a/ouster-ros/src/os_cloud_node.cpp
+++ b/ouster-ros/src/os_cloud_node.cpp
@@ -106,7 +106,8 @@ class OusterCloud : public OusterProcessingNodeBase {
                 info, tf_bcast.imu_frame_id(), timestamp_mode,
                 static_cast<int64_t>(ptp_utc_tai_offset * 1e+9));
             imu_packet_sub = create_subscription<PacketMsg>(
-                "imu_packets", selected_qos.keep_last(info.format.imu_packets_per_frame),
+                "imu_packets",
+                rclcpp::QoS(selected_qos).keep_last(info.format.imu_packets_per_frame),
                 [this](const PacketMsg::ConstSharedPtr msg) {
                     if (imu_packet_handler) {
                         // TODO[UN]: this is not ideal since we can't reuse the msg buffer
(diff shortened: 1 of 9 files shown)


matthew-lidar (commented):
This change looks good! Had to do a bit of research, but surprised the queue size is now a QoS setting with ROS2. Initially thought this was basically just history length (how many previous messages a subscriber would receive on joining) rather than the publish queue.

ROS2 really made something simple very confusing.

matthew-lidar (approved):
This change looks good! Had to do a bit of research, but surprised the queue size is now a QoS setting with ROS2. Initially thought this was basically just history length (how many previous messages a subscriber would receive on joining) rather than the publish queue.

ROS2 really made something simple very confusing.

ouster-ros/src/os_ros.cpp
```diff
@@ -50,6 +50,15 @@ size_t get_beams_count(const SensorInfo& info) {
     return info.beam_azimuth_angles.size();
 }
 
+rclcpp::QoS packet_qos(bool use_system_default_qos) {
+    // large enough to hold a full scan's worth of packets so a brief
+    // stall servicing the subscription callback doesn't drop packets
+    constexpr size_t packet_queue_depth = 1024;
```
matthew-lidar: Just a thought, would it make sense to calculate this based on how many packets are in a scan? Most scans are only 64 or 128 packets.

Samahu: With REV8/FW4.0 you can use have more if you choose the cols count to be less than 16.

Samahu: It is probably wise to estimate the exact value but I would have to parse the lidar mode and the cols count per packet which could take time.

Samahu: made dynamic

ouster-ros/src/os_cloud_node.cpp
```diff
@@ -106,7 +106,7 @@ class OusterCloud : public OusterProcessingNodeBase {
                 info, tf_bcast.imu_frame_id(), timestamp_mode,
                 static_cast<int64_t>(ptp_utc_tai_offset * 1e+9));
             imu_packet_sub = create_subscription<PacketMsg>(
-                "imu_packets", selected_qos,
+                "imu_packets", selected_qos.keep_last(info.format.imu_packets_per_frame),
```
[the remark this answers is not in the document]
Samahu: Ok, I didn't realize that this mutates the QoS in place

PR #575, opened 2026-09-09
Fix TIME_FROM_ROS_TIME stamping point clouds one frame period early

### LINKED ISSUE #541: Increased latency/pointcloud timestamps starting in version 0.14.0
**Describe the bug**
Starting with commit f56b1a0e0b12985c41d3f2630344fb5c5963f571, the timestamp in the pointcloud messages increases significantly, from ~53ms (slightly above one scan duration at 20Hz) to ~104ms. I have not verified if the actual "acquisition to message-receive" latency increases, or if this is just reporting the time wrong (could be accounting for scan duration twice?)

**To Reproduce**
1. Start ouster driver (we use `ouster_ros::OusterDriver` composable node) with configuration `lidar mode: 1024x20, lidar udp profile: RNG19_RFL8_SIG16_NIR16, imu udp profile: LEGACY`
1. Measure latency of `Pointcloud2` output with `ros2 topic delay /ouster/points`

Parameters:
```yaml
ouster/os_driver:
  ros__parameters:
    sensor_hostname: '192.168.141.190'
    udp_dest: ''
    mtp_dest: ''
    mtp_main: false
    lidar_mode: '1024x20'
    timestamp_mode: 'TIME_FROM_ROS_TIME'
    ptp_utc_tai_offset: -37.0
    udp_profile_lidar: ''
    metadata: ''
    lidar_port: 0
    imu_port: 0
    sensor_frame: lidar_link
    lidar_frame: os_lidar
    imu_frame: os_imu
    point_cloud_frame: lidar_link
    proc_mask: IMG|PCL|IMU|SCAN
    scan_ring: 0
    use_system_default_qos: false
    point_type: original
    auto_start: true
```

- Ouster Sensor: OS-0-32-U0
- Ouster Firmware Version: 3.1.0
- ROS version/distro: kilted
- Operating System: ubuntu 24.04
- Machine Architecture: x86
- First bad commit: f56b1a0e0b12985c41d3f2630344fb5c5963f571

eljasdistance:
## Related Issues & PRs

Fixes #541 (regression introduced by the count-based scan completion, first bad commit f56b1a0e as bisected in that issue).

## Summary of Changes

With `timestamp_mode: TIME_FROM_ROS_TIME`, every point cloud is stamped one full frame period early since 0.14.0 (+50 ms at 20 Hz, +100 ms at 10 Hz of *apparent* latency).

Mechanism: `lidar_handler_ros_time` stamps scan N with a value precomputed when scan N−1 completed, via `extrapolate_frame_ts(completing_packet)` = `receive_time − first_col_index × col_spacing`. That arithmetic assumed the completing packet is the **first packet of the next frame** (col index 0), which was true of the old `ScanBatcher`. Since the SDK finalizes a scan on its **own last packet** (`check_scan_complete` is packet-count based), the completing packet's first column index is `w − columns_per_packet`, so the "next scan start" lands one frame before the scan it is applied to.

This PR stamps the scan from its own data instead: the host arrival time of the scan's first received packet (`lidar_scan.packet_timestamp()`), walked back by `i × columns_per_packet × col_spacing` if the leading packet(s) were lost, with a warn-once fallback if no packet timestamps are present. The now-unused `lidar_handler_ros_time_frame_ts`, `extrapolate_frame_ts` and `packet_col_index` are removed.

Note the stamp semantics stay "host-clock time of scan start": the residual error is the delivery latency of the first packet (sub-millisecond on a quiet link), same as before 0.14.0.

## Validation

On a recorded drive (OS-1-128, 1024x20, `RNG19_RFL8_SIG16_NIR16`, fw 3.1, ouster-ros 0.14.2, `TIME_FROM_ROS_TIME`):

* header→receive hop of `/ouster/points` was 103.6 ms (p05 102.7 / p95 105.6, n=2000) against a structural expectation of ~54 ms (50 ms sweep + processing).
* The per-point `t` field spans exactly 49.95 ms, so the extra ~50 ms was not accumulation.
* Bug fingerprint: for 299/300 clouds, `header.stamp + 1008 × 48 828 ns` equals the host arrival timestamp of the **previous** frame's last packet **to the nanosecond** (cross-checked against `/ouster/telemetry`, which stamps `lidar_packet.host_timestamp` per packet). Packets arrive every 781 µs, so an accidental match is implausible.
* With this patch the expected hop is ~54 ms (sweep + ~4 ms processing/transport), matching the pre-0.14 behaviour reported in #541; the 49.22 ms correction is exactly the measured bias. Compile-verified on ROS 2 Jazzy (`colcon build --packages-up-to ouster_ros`, Release); a live re-measure on our rig is scheduled and I will post the number in this PR.
* `TIME_FROM_PTP_1588` and sensor-time paths untouched.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


nazneenkhan4:
I faced the same issue while testing, any reason for deleting this branch ? the given solution resolved the issue

